PR #27, opened 2014-02-11
Async

maksimkim:
Hi Jeremy,

I'd like FV very much. It helped me multiple times to decouple validation logic and business rules and keep code simpler. The only thing I've missed was support on async validation. E.g. to check Email field is unique one have to lookup the database in non-blocking way.

Please take a look at my pull request where I've tried to add this feature.

Changes description:
- All backbone interfaces like IValidator, IPropertyValidator and their derivatives now have asynchronous XXXAsync twin of Validate/IsValid method that returns Task<T>.
- I assume to have the following options to set async validators:
  - Through MustAsync overloads
  - Through CustomAsync overloads
  - By subclassing AsyncValidatorBase (which is PropertyValidator abstract descendant) and implementing IsValidAsync
- I've used TaskHelpers.Sources nuget package (which doesn't contain library, just two c# files) to simplify working with TPL.

### DIFF
--- a/src/FluentValidation.Tests/AbstractValidatorTester.cs
+++ b/src/FluentValidation.Tests/AbstractValidatorTester.cs
@@ -20,6 +20,7 @@ namespace FluentValidation.Tests {
 	using System;
 	using System.Globalization;
 	using System.Threading;
+	using System.Threading.Tasks;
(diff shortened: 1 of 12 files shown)


Korijn:
:+1:

PR #41, opened 2014-08-15
Fix test helper extension.

solo12zw74:
Fix 'ShouldHaveChildValidator' extension method. It works wrong if
SetValidator was called with some conditions ('When' condition)


solo12zw74:
Hello, Jeremy!

Thanx for response! Yes, i think it will be useful for test conditionally child validator

PR #42, opened 2014-09-04
Small change to support StopOnFirstFailure between rule chains

Tanhe:
So I'm not too familiar with this code base and my change feels a bit dirty (mostly don't like the naming similarity between the intra-chain cascade and the inter-chain cascade) but it clearly shows the functionality that I'm looking to achieve within FluentValidation. I find often that I'll have situations where the validation of one property requires that some other subset of properties already be proven valid. Or that the validation is non-trivial and I'd like it to bail as early as possible.

### DIFF
--- a/src/FluentValidation.Tests/CascadingFailuresTester.cs
+++ b/src/FluentValidation.Tests/CascadingFailuresTester.cs
@@ -119,5 +119,40 @@ namespace FluentValidation.Tests {
 			results.Errors.Count.ShouldEqual(1);
 		}
 
+	    [Test]
+	    public void ChainCascadeMode_StopOnFirstFailure_StopsChainProcessing() {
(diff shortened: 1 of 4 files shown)


quarion:
Personally, I would like to vote up for this pull request, unless this feature will be implemented differently.  
Having ordered validation suites, that will stop on first failure, is really helpful and can simplify complicated conditional validations.

JeremySkinner:
This won't be implemented, but I have now added support for explicit rule dependencies (see #65), which achieves the same result, but doesn't require an implicit ordering of rules.

Tanhe:
Thanks Jeremy. Haven't had much time to follow this but I'll check out the
new rule dependencies. Thanks!

~Ethan

PR #45, opened 2014-10-31
Added properties to Error that can be used to build customized message.

milanc:
Hi

Added ErrorCode property so certain actions can be taken inside business logic based on error code.
ErrorCode for built in validations are based on resource keys for messages e.g. "exact_length_error"
When WithLocalizedMessage is used then error code would be that resource message key.
Also some of private objects are exposed as properties that can help to build custom error message.

image

We are using validation in API, returned JSON can be used by  consumer to create custom maybe shorter, or more descriptive message for its end users.
This is current response based on our custom class (we have customized property names to be same as existing error object we had)

image

Hope this make sense.
Thanks
Milan

### DIFF
--- a/src/FluentValidation.Tests/ExactLengthValidatorTester.cs
+++ b/src/FluentValidation.Tests/ExactLengthValidatorTester.cs
@@ -16,99 +16,81 @@
 // The latest version of this file can be found at http://www.codeplex.com/FluentValidation
 #endregion
 
-namespace FluentValidation.Tests
-{
-    using System.Globalization;
-    using System.Linq;
-    using System.Threading;
-    using NUnit.Framework;
-    using Validators;
+namespace FluentValidation.Tests {
+	using System.Globalization;
+	using System.Linq;
+	using System.Threading;
+	using NUnit.Framework;
+	using Validators;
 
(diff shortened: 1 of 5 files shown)


di97mni:
Is it possible to set an ErrorCode for Custom as well?

```c#
public class MyCommandValidator : AbstractValidator<MyCommand>
{
    public MyCommandValidator()
    {
        Custom(c => ...);
    }
}
```

milanc:
As mentioned in initial comment, when WithLocalizedMessage is used then error code would be that resource message key. So error codes are setup during rule building.

I hope this answers your question.

di97mni:
I can't find `WithLocalizedMessage` on `Custom()`. How should it be used for Custom errors?

JeremySkinner:
Custom expects you to return the ValidationFailure objects directly - so anything for building custom messages is bypassed (WithLocalizedMessage etc). You can still set the properties on the ValidationFailure itself, but you basically have to hand-roll everything. 

I'd avoid using Custom unless you really need to. There are better ways to do custom validation - either using Must() or by using a Custom Property Validator. 

If you need further help, please open a separate issue.

PR #48, opened 2015-03-25
Fixed problem with WebApi and collections.

calvinb:
When validating a model with a collection property via WebApi, if an error is found in the collection, the other validators on the parent model are ignored. I have added a unit test for this case and made a minor adjustment to keep from short-circuiting the other validators.


JeremySkinner:
That's great, thanks!

PR #59, opened 2015-04-28
Add a rule remove method

HansHuang:
When reuse the validator by inheritance, we may want to remove some existing rules inherited from parent.

### DIFF
--- a/src/FluentValidation/AbstractValidator.cs
+++ b/src/FluentValidation/AbstractValidator.cs
@@ -261,7 +261,23 @@ namespace FluentValidation {
 			When(x => !predicate(x), action);
 		}
 
-		/// <summary>
+	    /// <summary>
+	    /// Remove a validation rule for a specify property from existing.
+	    /// </summary>
+	    /// <example>
+	    /// RemoveRule(x => x.Surname)...
+	    /// </example>
+	    /// <typeparam name="TProperty">The type of property being validated</typeparam>
+	    /// <param name="expression">The expression representing the property to validate</param>
+	    /// <returns>Successed removed or not</returns>
+	    public bool RemoveRule<TProperty>(Expression<Func<T, TProperty>> expression) {
+	        expression.Guard("Cannot pass null to RemoveRule");
+	        var member = expression.GetMember();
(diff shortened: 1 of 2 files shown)


pashute:
Can a ruleset share previously defined rules with another ruleset? i.e. Can one rule be part of two rulesets?  If so, you don't need to remove the rules, just to put them in your new ruleset not inherited from the parent.
 
Also: How would you know which rule to remove, where do you see your rules besides in the code. 
They are hidden from you at design time. No?

HansHuang:
Updated my last comment that may cause some misunderstanding. Sorry for that. ;) Yes, one rule can't be part of two rulesets.
Let's suppose a scenario: the base class not allow PropertyA be a nagative number, but the child class make some enhancement that allow nagative for PropertyA; so for the validaton of child class, step first of course is inherit the validation of base class, and next is want to remove the validation rule of PropertyA.
Yes, all the specific rules are hidden at design time, but we may remove all the rule base on the property. This is not very proper. And I should modify my code.
Do u have any good solution for our scenario :)

JeremySkinner:
I'm not particularly keen on this, as it kinda breaks the immutability of validators and adds the potential for confusion if rules are being removed or modified at runtime.

Personally I don't like validator inheritance at all - I'd much rather have 2 distinct validators for validating your two classes, each with their own set of rules. Yes, this means that there will be some duplication of code, but this can be minimized with extension methods or custom property validators. If I have a choice between code clarity and code reuse, I always pick clarity.

Also, what if you have multiple rules defined for the same property? How do you know which one should be removed? Your code will remove the first, which I'm not sure is the best approach - it'll also fail if there isn't a rule specified for the property. 

Let me have a think about whether there's a nicer way of doing this...it might be that we could have some sort of "rule override" mechanism. I'll leave this issue open for now until I've had a chance to investigate it.

Jeremy

Raknison123:
Yes, I would also appreciate a RuleSet override mechanism.
In my case I have kind of a BaseValidator class with some "virtual" RuleSets.
In my child classes I want to override those "virtual" RuleSets from the Base validator class.

PR #79, opened 2015-06-05
Pass validator provider to SetValidator

roend83:
I added the ability to pass a validator provider to set validator so the child validator has access to the parent object.  Similar functionality already exists for SetCollectionValidator.


JeremySkinner:
Thanks - this looks good. Please could you format indentations with tabs rather than spaces for consistency with the rest of the project?

roend83:
Done

PR #93, opened 2015-07-09
Added lambdas on LengthValidator and RegularExpressionValidator to ac…

### LINKED ISSUE #92: Access to validator instance properties when using property validators in a rule
Hello,

When I define an AbstractValidator for a class, I want to be able to access my class properties as parameters for the property validators the same way I can pick the property to validate when calling RuleFor()

Example:
```c#
public class TextFieldValidator : AbstractValidator<TextField>
    {
        public TextFieldValidator()
        {
            When(t => !string.IsNullOrEmpty(t.Regex), () =>
            {
                // This doesn't work
                RuleFor(t => t.Value).Matches(t => t.Regex); 
            });
        }
    }
```

There is a workaround to do something like that?
Is it planned to add that feature in a future version?

emisand:
Added lambdas on LengthValidator and RegularExpressionValidator to access properties from the validator instance. This closes #92

### DIFF
--- a/src/FluentValidation.Tests/LengthValidatorTests.cs
+++ b/src/FluentValidation.Tests/LengthValidatorTests.cs
@@ -39,9 +39,25 @@ namespace FluentValidation.Tests {
 		}
 
 		[Fact]
+		public void When_the_text_is_between_the_lambda_range_specified_then_the_validator_should_pass()
+		{
+			var validator = new TestValidator(v => v.RuleFor(x => x.Surname).Length(x => x.MinLength, x => x.MaxLength));
+			var result = validator.Validate(new Person { Surname = "Test", MinLength = 1, MaxLength = 10 });
+			result.IsValid.ShouldBeTrue();
+		}
+
+		[Fact]
 		public void When_the_text_is_smaller_than_the_range_then_the_validator_should_fail() {
 			var validator = new TestValidator(v => v.RuleFor(x => x.Surname).Length(5, 10));
 			var result = validator.Validate(new Person { Surname = "Test" });
+			result.IsValid.ShouldBeFalse();
+		}
+
+		[Fact]
+		public void When_the_text_is_smaller_than_the_lambda_range_then_the_validator_should_fail()
+		{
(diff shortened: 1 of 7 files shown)


JeremySkinner:
Thanks for this. A few comments:

- You should use Func<T, int> rather than Expression<Func<T, int>> etc so you don't need to compile the expression. You'd only need to use an expression if you need to get metadata about the lambda, but in this case it isn't neccessary - we just need the delegate itself

- I don't think the casts to int in LengthValidator line 58 are neccessary

- Please could you use tabs rather than spaces for indentation for consistency with the rest of the project

- The changes to Regex validator on line 39 mean that you can no longer pass in a pre-configured RegEx instance. This behaviour needs to be maintained - I'd suggest storing it inside the lambda: `this.regexFunc = x => regex;`  

- Please could you add appropriate unit tests for the new methods 

If you can make these changes then I'll merge this in.

Thanks!

emisand:
I have made all the changes you noted.

Please note that when Lambdas are used for the modified validators, properties like Min or Max in the LengthValidator or Expression in RegularExpressionValidator may not be accesible before IsValid is called, because lambdas are excecuted inside IsValid. Let me know if this is an acceptable behavior.

Thank you.

JeremySkinner:
Thanks - this looks good. I'll try and merge it later today. 

I think the behaviour for Min/Max etc is fine, but it may be worth introducing an extra property on the validators so you can check whether or not they're accessible for metadata purposes (IsMetadataReady or something like that). I'll have a think about this.

JeremySkinner:
I've merged this into the latest commit. Thanks again.

emisand:
I've found a problem with this that we shold think how to fix.

When using this new lambdas for fluent validation with the MVC5 project, the client validation rules are not generated correctly because the code that generates this client validation rules does not use the ValidationContext to evaluate the lambdas.

What do you recommend me to evaluate the lambdas and fill the validation properties before the client validation rules are generated?
If you have any ideas, I will try to fix it and submit a pull request.

Thank you.

JeremySkinner:
There isn't any contextual information when generating client-side rules, so anything that is cross-property won't work client-side. I'd suggest modifying the clientside adaptors to explicitly skip trying to generate client-side rules when using the lambda-based overloads, and assume they are server-side only.

PR #114, opened 2015-09-02
Support simple chinese

leehom0123:
Add resource file to support simple chinese


JeremySkinner:
I've merged this in - thanks!

PR #118, opened 2015-09-10
Polish resources update

gralin:
I have added two missing entries in polish resources and corrected one that was wrong. In 'exact_length_error' the value was 'Maximal field lenght is X characters. Y characters were entered.'


JeremySkinner:
Thanks!

Issue #122, opened 2015-09-16
Empty and Null validators


regisbsb:
I'm finding myself to have to implement the Empty and Null validators.
@JeremySkinner  Is there scope to add that to the library?

Cheers.

JeremySkinner:
Sure, if you want to submit a pull request I'd happily include them.

regisbsb:
Cool I will. Cheers.

regisbsb:
@JeremySkinner  Done!

PR #124, opened 2015-09-16
Implemented #122 Empty and Null validators

regisbsb:
Please double check if it's only that is necessary or do I need to modify more files to properly add them.

Cheers
#122


JeremySkinner:
Thanks for this...could you add some corresponding unit tests too? Thanks

regisbsb:
@JeremySkinner unit tests added!

JeremySkinner:
Looks good...I'll merge this in, but I noticed the default messages were changed for NotNull. I'm not really keen on this - can we put them back to how it was? If we want to change it, we'd also need to change all the other translations, but I actually prefer it to say "empty" rather than "null"

regisbsb:
Agreed. I'll follow the same logic then.

regisbsb:
Done.

JeremySkinner:
Thanks, I've merged it in

Issue #125, opened 2015-09-18
Collection indexing assumes collections are int indexed


audunm:
If I have an item in my MVC modelstate with key `ProductConfigurationAlternatives[24b1f963-713b-4837-8e19-d565233ba677].ProductConfigurationAlternativeItems[b1d22400-0213-447b-9974-0653151d2b5e].Id`, and if it fails validation, the error is not added with the correct key. Instead it is added with the key `ProductConfigurationAlternatives[0].ProductConfigurationAlternativeItems[1].Id`

My collection is guid-indexed, but fluent validation assumes it's integer indexed.

JeremySkinner:
That's not something that's supported I'm afraid...you'd need to write a custom implementation of the ChildCollectionValidator if you need support for anything other than integer-based indicies.

SeanKilleen:
Cleanup: Closing this as it hasn't been active in some time. Feel free to re-open if necessary.

PR #126, opened 2015-09-18
Added support for correct indexes in non-integer indexed childcollections

revioaudun:
This will allow users to implement interface IIndexedCollectionItem for child collections items that uses something else than integers as indicies. Makes FluentValidation work well with BeginCollectionItem-helper in MVC


revioaudun:
This will solve Issue #125

JeremySkinner:
Thanks!

PR #187, opened 2016-01-08
Added an optional RuleSet arguments.

PureKrome:
- Updated method signature for ValidateAndThrow and ValidateAndThrowAsync.

### DIFF
--- a/src/FluentValidation.Tests/ValidateAndThrowTester.cs
+++ b/src/FluentValidation.Tests/ValidateAndThrowTester.cs
@@ -38,7 +38,21 @@ namespace FluentValidation.Tests {
 			typeof(ValidationException).ShouldBeThrownBy(() => validator.ValidateAndThrow(new Person()));
 		}
 
-		[Fact]
+        [Fact]
+        public void Throws_exception_with_a_ruleset() {
+            var validator = new TestValidator {
+                v => v.RuleFor(x => x.Surname).NotNull()
+            };
+
+            const string ruleSetName = "blah";
+            validator.RuleSet(ruleSetName, () => {
+                validator.RuleFor(x => x.Forename).NotNull();
+            });
+
+            typeof(ValidationException).ShouldBeThrownBy(() => validator.ValidateAndThrow(new Person(), ruleSetName));
+        }
+
+        [Fact]
 		public void Throws_exception_async() {
 			var validator = new TestValidator {
 				v => v.RuleFor(x => x.Surname).NotNull()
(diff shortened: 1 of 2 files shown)


src/FluentValidation.Tests/TestValidator.cs
```diff
@@ -30,4 +30,14 @@ public class TestValidator : InlineValidator<Person> {
 			}
 		}
 	}
+
+    public class TestValidatorWithRuleSet : InlineValidator<Person>
```
PureKrome: I'm cringing at adding this class in because I can't figure out how to add a `RuleSet` to the normal `TestValidator`.

src/FluentValidation.Tests/ValidateAndThrowTester.cs
```diff
@@ -38,7 +38,16 @@ public class ValidateAndThrowTester {
 			typeof(ValidationException).ShouldBeThrownBy(() => validator.ValidateAndThrow(new Person()));
 		}
 
-		[Fact]
+        [Fact]
+        public void Throws_exception_with_a_ruleset() {
+            var validator = new TestValidatorWithRuleSet {
```
PureKrome: I'd prefer to use `TestValidatorWithRuleSet` instead of `TestValidatorWithRuleSet` and inline add the `RuleSet`. 

Do you know how I could do that (so I can update this file) ?

PureKrome:
Now to try and get this pushed up as quickly as possible in a sneaky build, plz @JeremySkinner

JeremySkinner:
Merged in latest commit- thanks!

PureKrome:
!! W O O T !!

PureKrome:
BTW - when will this be available on nuget? soonish?

JeremySkinner:
Hopefully later this week or next

justsayno:
Awesome work @PureKrome I'm also really keen for this to get on Nuget, This library does exactly what I need with the addition of this parameter. 

How would I use the latest version of this without Nuget? I can't figure out what I need to do to get that to work.

PureKrome:
@justsayno - if you cannot wait for the new nuget to be published, then you'll need to download the source, compile it, grab the `.dll

justsayno:
@PureKrome Thanks, I was not on the scene in the Bad-Old-Days so this is first for me!

justsayno:
I just did what you said above and I still don't have that overload... Also when I ran it the `Build.cmd` the test project said it could not be found so I commented out that part of the build...

Any ideas why I would be having issues? The version it thinks it is is 1.0.0.... In VS2015.

Just to explain what I did:

Cloned repo, ran the Build.cmd. Then copied the FluentAssertion folder from build to my project and referenced it from the project I am using it.

PureKrome:
Hmm.

1. Checking the [commits](https://github.com/JeremySkinner/FluentValidation/commits/master) table for the `master` branch, my PR/commit is in there (scroll down to Jan 8th 2016).
2. Check what branch you're using. I'm assuming it's `master` which has my commit.
3. Have you gotten the _latest_ code/commits from GH? (ie, do a PULL)
4. What version of VS do you have installed?

JeremySkinner:
Please don't use the VS2015 project in the master branch- this is for the aspnet vnext integration. You'll need to use VS2013 solution for now.

JeremySkinner:
Until yesterday, the VS2015 project was only for asp.net vnext integration. I recently committed a change that now uses VS2015 for the main solution. Please make sure you've pulled down the latest code.

justsayno:
Ahh so sorry guys I just doing something else wrong and completely misread this thread! Works fine as is. Sorry for the confusion!

JeremySkinner:
@PureKrome I've just uploaded a 6.1 package. Please give it a try and let me know if it looks OK.

PR #214, opened 2016-02-22
Enum validator

Lechus:
Validate if enum value is defined in Enum.
Credits to @Bobby B from stackoverflow.com

### DIFF
--- a/src/FluentValidation.Tests/EnumValidatorTests.cs
+++ b/src/FluentValidation.Tests/EnumValidatorTests.cs
@@ -1,0 +1,44 @@
+namespace FluentValidation.Tests
+{
+    using System.Linq;
+    using Xunit;
+
+    public class EnumValidatorTests {
+		TestValidator validator;
+
+		public EnumValidatorTests() {
+			CultureScope.SetDefaultCulture();
+
+			validator = new TestValidator {
+				v => v.RuleFor(x => x.Gender).IsInEnum()
+			};
+		}
+
+		[Fact]
+		public void IsValidTests() {
+			validator.Validate(new Person { Gender = EnumGender.Female }).IsValid.ShouldBeTrue();  // Simplest valid value
+			validator.Validate(new Person { Gender = EnumGender.Male }).IsValid.ShouldBeTrue();    // Other valid value
+            validator.Validate(new Person { Gender = (EnumGender)1 }).IsValid.ShouldBeTrue();      // Casting with valid value
+        }
+
+        [Fact]
+        public void When_the_enum_is_not_initialized_with_valid_value_then_the_validator_should_fail() {
+            var result = validator.Validate(new Person());                                         // Default value 0 is not defined in Enum
+            result.IsValid.ShouldBeFalse();
+        }
+
+        [Fact]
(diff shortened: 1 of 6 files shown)


Lechus:
@JeremySkinner It does not work with Nullable Enum.
Quick fix could be something like:

	
	protected override bool IsValid(PropertyValidatorContext context) {
			
            //Nullable Enum
            Type enumType = Nullable.GetUnderlyingType(_enumType) ?? _enumType;
            
            if (!enumType.IsEnum() || null == context.PropertyValue) return false;

            return Enum.IsDefined(enumType, context.PropertyValue);
		}

JeremySkinner:
Good catch....I'll make the change to check the underlying type.

However, I'm not really happy with the validator failing if the property value is null....this isn't really consistent with the rest of FluentValidation- the other validators typically treat null as valid, and you have to explicitly chain the NotNull() validator if you want nulls to fail validation. 

Is that OK?

Lechus:
I added to `
Person {
  public EnumGender? NullableGender { get; set; }
}
`

Tests updated with:
```
           nullableValidator = new TestValidator {
                v => v.RuleFor(x => x.NullableGender).NotNull().IsInEnum()
            };
```

Test:

       [Fact]
        public void When_the_Nullableenum_is_not_initialized_with_valid_value_then_the_validator_should_fail()
        {
            var result = nullableValidator.Validate(new Person());
            result.IsValid.ShouldBeFalse();
        }

Test fails because Enum.IsDefined() throws Exception: System.ArgumentNullException.
That's why there is the validator failing if the property value is null.

Lechus:
Test will pass if I StopOnFirstFailure:
```
            nullableValidator = new TestValidator {
                v => v.RuleFor(x => x.NullableGender).Cascade(CascadeMode.StopOnFirstFailure).NotNull().IsInEnum()
            };
```

but I bet You know better solution.

JeremySkinner:
No need really for a test in that particular case, as it's actually testing the NotNull rather than the enumvalidator. 

Here are the tests I've added:

```
[Fact]
public void Nullable_enum_valid_when_property_value_is_null()
{
    var validator = new InlineValidator<Foo>();
        validator.RuleFor(x => x.Gender).IsInEnum();
    var result = validator.Validate(new Foo());
}

[Fact]
public void Nullable_enum_valid_when_value_specified()
{
    var validator = new InlineValidator<Foo>();
    validator.RuleFor(x => x.Gender).IsInEnum();
    var result = validator.Validate(new Foo() { Gender = EnumGender.Male });
}

[Fact]
public void Nullable_enum_invalid_when_bad_value_specified()
{
    var validator = new InlineValidator<Foo>();
    validator.RuleFor(x => x.Gender).IsInEnum();
    var result = validator.Validate(new Foo() { Gender = (EnumGender)42 });
    result.IsValid.ShouldBeFalse();
}

public class Foo
{
    public EnumGender? Gender { get; set; }
}
```

JeremySkinner:
I also just noticed that the error message for EnumValidator isn't localised like the others. I'll get that fixed too.

Lechus:
I see now. Cool. Let it be consistent!
Cheers!

JeremySkinner:
I've pushed this out as v6.2.1

Lechus:
Thank you.

Issue #222, opened 2016-03-01
EnumValidator: [Flags] Support


broccliman:
Nice work with the new `IsInEnum()` functionality:

 - https://github.com/JeremySkinner/FluentValidation/commit/0a9fb88978a5912865a27671a1b4d079ae476ee5
 - https://github.com/JeremySkinner/FluentValidation/commit/556ccde11de362e0b8b3b141b757adf4155d2a7d

I've had issues in the past using `Enum.IsDefined` for validating enum input where the enum uses the `FlagsAttribute`, and it looks like 6.2.1 would fail for the following scenario:

```c#
public class Foo
{
    public FooEnum EnumValue { get; set; }
}

[Flags]
public enum FooEnum
{
    A = 0,
    B = 1 << 0,
    C = 1 << 1,
}

var validator = new InlineValidator<Foo>();
validator.RuleFor(x => x.EnumValue).IsInEnum();

var isValid = validator.Validate(new Foo { EnumValue = FooEnum.B | FooEnum.C }).IsValid;
Assert.IsTrue(isValid);
```

Lechus:
@broccliman good catch.  If you want you can create a pull request with new tests, amend IsDefined to check for flags. @JeremySkinner can then review and merge this support for Enums with flags,

broccliman:
My PCL knowledge is limited, but informed enough to know that type related apis/reflection varies across each platform. I've included an implementation below that works for the full framework, but will need some tweaks to be compatible across the platforms that FV targets. I'd rather defer to those more familiar with FV's goals around compatibility/perf/mem/culture/.. some of the explicitness in my implementation could be simplified in exchange for less code + more memory allocation. As an example, [an answer on this SO thread suggests to use ToString() and check the first character](http://stackoverflow.com/questions/527486/c-enum-isdefined-on-combined-flags). I'd imagine that implementation would have no issues with PCL compatibility, but there may be some nuances around culture to consider (also not my strong suit).

I hope the bug report, example test case failure, and example implementation below were helpful. I wish I was more in tune with FV project goals to wrap this up nicely in a PR.

```c#
namespace FluentValidation.Validators
{
    using System;
    using System.Collections.Generic;
    using FluentValidation.Internal;
    using Resources;

    public class EnumValidator : PropertyValidator
    {
        private readonly Type _enumType;

        public EnumValidator(Type enumType) : base(() => Messages.enum_error)
        {
            this._enumType = enumType;
        }

[...]

PR #223, opened 2016-03-02
Add support for [Flags] enumerations in EnumValidator.

broccliman:
- Misc change: add missing file headers to EnumValidator and EnumValidatorTest
- Misc change: make validation message in resource file more descriptive
>> Including a user-provided value as part of a system's output is potentially vulnerable to injection style attacks; however, since out-of-range enum values will only contain numeric values, its safe to include {PropertyValue} as part of the validation error message.

---
In response to: #222 

@JeremySkinner .. I did my best to maintain style with the theme of the repo (tabs, curly-brace, ..). Also, the implementation I used is somewhat verbose, but the intent was to minimize unnecessary object allocations.

@Lechus .. I changed the validation message that you included in your original commit. I think the additional information will be useful (and safe) to someone who isn't aware that out-of-range numeric values can be used for a corresponding enum member.

### DIFF
--- a/src/FluentValidation/Internal/Compatibility.cs
+++ b/src/FluentValidation/Internal/Compatibility.cs
@@ -61,13 +61,19 @@ namespace FluentValidation.Internal
 #endif
 		}
 
-		public static ValidatorAttribute GetValidatorAttribute(this Type type)
+		public static TAttribute GetCustomAttributeCompatible<TAttribute>(this Type type)
+			where TAttribute : Attribute
 		{
 #if PORTABLE || CoreCLR
-			return type.GetTypeInfo().GetCustomAttribute<ValidatorAttribute>(true);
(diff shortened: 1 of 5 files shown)


src/FluentValidation/Internal/Compatibility.cs
```diff
@@ -61,15 +61,21 @@ public static Func<string> CreateGetter(this PropertyInfo property)
 #endif
 		}
 
-		public static ValidatorAttribute GetValidatorAttribute(this Type type)
+		public static TAttribute GetCustomAttributeCompatible<TAttribute>(this Type type)
```
broccliman: @JeremySkinner .. I needed this to check for the existence of `FlagsAttribute`. I thought it would be better to expose this utility and encapsulate the portability aspects, rather than let that leak into `EnumValidator`.

JeremySkinner:
Thanks- that's merged in.

PR #239, opened 2016-04-12
Null Extension has the correct xml doc

erendrake:


src/FluentValidation/DefaultValidatorExtensions.cs
```diff
@@ -45,8 +45,8 @@ public static class DefaultValidatorExtensions {
 		}
 
         /// <summary>
-		/// Defines a 'not null' validator on the current rule builder. 
-		/// Validation will fail if the property is null.
+		/// Defines a 'null' validator on the current rule builder. 
+		/// Validation will fail if the property is not null.
 		/// </summary>
```
erendrake: This appears to be a simple cut and paste error from the NotNull Extension

JeremySkinner:
Already fixed

Issue #98, opened 2015-07-30
vNext / MVC 6 Integration


endeffects:
Hi,

i saw you have created a branch for vNext / Mvc6. When do you plan a beta / final release?

Regards

JeremySkinner:
I am away with work until September, so I'm not going to have much chance to work on it until then I'm afraid. I expect mid/end September will be the earliest I can do an initial release (unless someone else wants to help out)

endeffects:
Thanks for your quick response. Than i have tu put this into my backlog. :)

ddotlic:
@JeremySkinner I might have some time to do this at work (win-win for everyone) since my employer needs this (we're starting our work on next major version of the product and we'll be depending on Fluent Validation).

Is the work to be done relatively easy to explain? The hooks in MVC 6 should be much cleaner than before, I recall a brief exchange with you a few months back happened exactly at the moment Microsoft finally responded to one of your ASP.NET MVC github issues on the subject of Fluent's integration with MVC6, you were very hopeful following that.

What can I do to help?

JeremySkinner:
Thanks, I appreciate the offer. At the moment there's not much more that can be done...I'm still waiting for the asp.net team to make the following changes:

- Support for localization satellite assemblies in vnext projects
- Changes to the validation API to make it easier to add 3rd party validation libraries

Once these are done then I can do the following:

- Start using the vnext project file to generate outputs for all supported frameworks (rather than having separate projects for .net, vnext, pcl etc). I've done most of the work for this already, I just need to re-enable it 
- Add the MVC integration project. This is where most of the work will be - I haven't started on it at all because the API is still in flux. I don't quite know what it'll look like yet until I see the final validation API in MVC, but it should be quite different to the existing integration projects. 

I'll post an update here when I know more about what'll be needed.

ddotlic:
> Changes to the validation API to make it easier to add 3rd party validation libraries

I was under (did not check) impression they've done it based on your feedback? Still not done? Sigh. It's quite late to be adding these kind of modifications (they say they'll be feature complete in the next beta which is due in 2-3 weeks time).

>Start using the vnext project file

Based on what I've seen elsewhere this is going to be a pain.

>Add the MVC integration project. This is where most of the work will be - I haven't started on it at all because the API is still in flux.

Shame that the API still hasn't stabilized.

Oh well, let's revisit this subject a bit later :smile:

ddotlic:
[...]

PR #263, opened 2016-05-23
Add NetStandard support

jeremymeng:
For part of #98. 

This add a new PCL project for netstandard1.0 support and a new unit test project that targets netcoreapp1.0.  The new test project requires the .NET Core SDK to run.


JeremySkinner:
Thanks! I'll try and get this merged in this week. Couple of questions:

- Are there any issues from targeting netstandard1.0? I thought we needed to be on 1.3?
- What's the purpose of the FluentValidation.xproj?

jeremymeng:
> Are there any issues from targeting netstandard1.0? I thought we needed to be on 1.3?

Here's a quote from https://github.com/dotnet/corefx/blob/master/Documentation/architecture/net-platform-standard.md#mapping-the-net-platform-standard-to-platforms.

"In general, class libraries which target a lower .NET Platform Standard version, like 1.0, can be loaded by the largest number of existing platforms, but will have access to a smaller set of APIs. On the other hand, class libraries which target a higher .NET Platform Standard version, like 1.3, can be loaded by a smaller number of newer platforms, but will have access to a larger, more recent set of APIs."

Since this port is basically based on PCL profile259, it can support netstandard1.0 and be consumed by higher netstandard versions.  Unless newer .NET Core APIs are used in cross-compilation on NET45/NET46/higher (the only conditional other than `PORTABLE` is `WIN10` in the NET46 project for using `Contract`, which is in netstandard1.0), I think we could benefit more by targeting as low as possible.

> What's the purpose of the FluentValidation.xproj?

It is a file created by Visual Studio tooling, and served as some kind of wrapper for project.json to interact with MSBuild. VS creates it when you create a project from a .NET Core/ASP.NET Core project template.  VS also creates it if you add an existing project to the solution and point to a `project.json` file.

PR #275, opened 2016-06-09
AspNetCore MVC integration

cleftheris:
I have added a new project FluentValidation.MvcCore. The integration project is working perfectly but in order to build it I had to change the `global.json` projects import Not to include the wrapper xproj. That brakes the NetStandard tests project. Maybe @jeremymeng could check it out.

Also the itegration with an MVC project is as easy as 

``` csharp
using FluentValidation.MvcCore;

public class Startup {
   ...

   public void ConfigureServices(IServiceCollection services) {

            services.AddMvc()
                    .AddFluentValidation(Assembly.GetEntryAssembly());
   }
}

```

Last but not least the validation plumbing in Mvc core is giving an extra option to mark a validator as `Reusable` upon discovery (check [FluentValidationModelValidatorProvider.cs](https://github.com/cleftheris/FluentValidation/blob/1e138c769d8e999cf15a9480849615c9f8d7deb7/src/FluentValidation.MvcCore/FluentValidationModelValidatorProvider.cs#L27)). This flag makes validators although registered as Transient with the DI to be treated (or not) as singletons. So I beleave that a custom `Attribute` or marker interface or base class could be used to trigger the flag on/off per validator.

This is related with #98 and could potentially help close it.

### DIFF
--- a/FluentValidation.sln
+++ b/FluentValidation.sln
@@ -7,7 +7,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution 
 	ProjectSection(SolutionItems) = preProject
 		Changelog.txt = Changelog.txt
 		FluentValidation.snk = FluentValidation.snk
-		src\FluentValidation.Tests-vnext\global.json = src\FluentValidation.Tests-vnext\global.json
+		global.json = global.json
 	EndProjectSection
 EndProject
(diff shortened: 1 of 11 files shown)


JeremySkinner:
Thanks, but this seems to be using mostly the same approach for MVC integration as the MVC5 version (with a ValidatorProvider)- the aim for the MVC6 integration is to switch over to using IObjectModelValidator. Please see this post which documents the roadmap:

https://github.com/JeremySkinner/FluentValidation/issues/98#issuecomment-219080143

cleftheris:
Yes you are correct I based it on what was already in there. Just wanted to share maybe the part in MvcBuilderExtension is worth looking regarding your point on previous integration mechanisms like `FluentValidationModelValidatorProvider.Configure`. 

Also thanks for taking the time to check it out :-)

PR #287, opened 2016-06-23
Add support for length validation message grammar

billbogaiv:
Previous behavior uses fixed grammar and does not make sense with certain numerical combinations (i.e. 1 => characters vs. 1 => character).

Only provided English-version of resource file. There is no associated issue with this PR. It is just something I encountered while using the library.

### DIFF
--- a/src/FluentValidation/Validators/LengthValidator.cs
+++ b/src/FluentValidation/Validators/LengthValidator.cs
@@ -65,7 +65,16 @@ namespace FluentValidation.Validators {
 				context.MessageFormatter
 					.AppendArgument("MinLength", Min)
 					.AppendArgument("MaxLength", Max)
-					.AppendArgument("TotalLength", length);
+					.AppendArgument("TotalLength", length)
+					.AppendArgument("MinLengthCharacter", Min == 1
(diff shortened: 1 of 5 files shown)


JeremySkinner:
Thanks for this- I'm not very comfortable with making a change like this to just the English translation of the errors as it creates an inconsistency between the different resource files. Let me have a think about this over the weekend.

src/FluentValidation/Validators/LengthValidator.cs
```diff
@@ -65,7 +65,16 @@ public LengthValidator(Func<object, int> min, Func<object, int> max)
 				context.MessageFormatter
 					.AppendArgument("MinLength", Min)
 					.AppendArgument("MaxLength", Max)
-					.AppendArgument("TotalLength", length);
+					.AppendArgument("TotalLength", length)
+					.AppendArgument("MinLengthCharacter", Min == 1
```
billboga: This argument is not currently used in any message and was added to complete the set of available placeholders in the resource files.

JeremySkinner:
Cleanup old pull request; not implementing due to inconsistencies created with other translations.

PR #298, opened 2016-07-14
Tweaks to the FluentValidation solution

agehrke:
Hi

This pull request contains two commits.

1. Define official NuGet package source in NuGet.config, instead of falling back to what's configured in Visual Studio.
2. Make use of [EditorConfig](http://editorconfig.org/) to define indentation rules. EditorConfig are used in many other OSS projects, including CoreCLR.

Andreas

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,9 @@
+# top-most EditorConfig file
+root = true
+
+[*.cs]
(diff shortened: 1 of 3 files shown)


JeremySkinner:
Looks good, thanks!

PR #313, opened 2016-08-19
Allow to register validators from multiple assemblies in AspNetCore.

nvivo:
Currently, when doing something like:

```cs
public void ConfigureServices(IServiceCollection services)
{
    services
        .AddMvc()
        .AddFluentValidation(config =>
        {
            config.RegisterValidatorsFromAssemblyContaining<ClassInSharedProject>();
            config.RegisterValidatorsFromAssemblyContaining<ClassInLocalProject>();
        });
}
```

Only the last assembly is registered. This PR allows for multiple assemblies to be registered.

Note: for some reason, the FluentValidation.dll sources are not compiling in my machine, so I wasn't able to properly test this, but I didn't want to just leave issue for this.


JeremySkinner:
Thanks!

PR #319, opened 2016-09-01
Support a pluggable join separator in the property chain.

RyanThomas73:
…ToString() implementation to use ValidatorOptions.PropertyChainSepartor to allow pluggable support of different join separators.


JeremySkinner:
Thanks!

Issue #331, opened 2016-10-01
Performance improvements to name resolution


batzen:
Hi,

i had to notice that name resolution for PropertyRule takes a considerable amount of time.

During our automated UI tests some validators/rules are called millions of times.

I got a modified version of FluenValidation to which i added caching.

My test:
- Validator consisting of two rules on different properties
- Running Validate 1 000 000 times

My results:
- No caching: 12 seconds
- Caching: 6 seconds

My question is:
Would you accept a PR adding said caching?

JeremySkinner:
Yes that sounds very useful. If you'd be happy to submit a pull request
I'll happily review it. It might not be for a while though as I'm very busy
with other commitments at the moment, but I'd definitely like to get more
performance improvements in the project.

batzen:
Ok, i will add a list of things i'd like to change to the issue itself and update it as i find more places that could be improved in regards of performance.
When i am finished with my list i will mention you here so that you can decide which things you'd like to be included in my PR.

One more question, do you prefer one PR or multiple PRs for each change?

batzen:
With all my optimizations in place i currently get the following results using BenchmarkDotNet on my machine:

| Name | Median | StdDev
|---|---|---|
| Version from nuget (6.2.1.0) | 7.6984 us | 0.0473 us
| My version | 1.9282 us | 0.0438 us

I used the following validator:
```
new TestValidator
	{
		v => v.RuleFor(x => x.Address).NotNull(),
		v => v.RuleFor(x => x.Address.Country).NotNull().When(x => x.Address != null)
	};
```

The benchmark only calls `validator.Validate(person);` where person is the Person class from the FluentValidation.Test project.
The validator and person instances are only created once during the benchmark.

Performance could be further improved if there was a way to cache the strings obtained from the resources, but that's not on my list as i only used FluentValidation with a custom provided ResourceProvider which is not localized (my application is only provided in german).

On my machine calling BuildErrorMessage 1 000 000 times takes about 6 seconds.
2 of these 6 seconds account for GetString on LocalizedStringSource.

JeremySkinner:
Looks good.

Regarding pull requests, it really depends on the size. If it's a huge
number of changes then several smaller ones might be better. I'll leave it
up to you to decide what you think's most appropriate.

Thanks!

batzen:
Changing SplitPascalCase from using Regex to a specialized method using a StringBuilder i get the following results. 
[...]

PR #334, opened 2016-10-05
Various performance improvements during validation

batzen:
This PR primarily focuses on caching and skipping/delaying things done during validation.
Especially the creation of validation errors.

In my tests (some posted in #331) these changes brought about 6 times better performance and cut memory traffic in half.
Used a validator with just two rules in my tests.

### DIFF
--- a/.build/build.proj
+++ b/.build/build.proj
@@ -22,6 +22,10 @@
 	<Target Name="Test">
     <MSBuild Projects="test.proj" Properties="Configuration=$(Configuration)" />
 	</Target>
+	
+	<Target Name="TestMvcCore">
+    <MSBuild Projects="test.proj" Properties="Configuration=$(Configuration)" Targets="TestMvcCore" />
+	</Target>
 
   <Target Name="Deploy">
     <RemoveDir Directories="$(BuildDir)" />
--- a/.build/test.proj
+++ b/.build/test.proj
@@ -22,8 +22,12 @@
 
   <Target Name="default">
     <xunit Assemblies="@(TestAssemblies)" ParallelizeAssemblies="false" ParallelizeTestCollections="false" />
-    <exec Command="dotnet test --no-build" WorkingDirectory="$(BaseDir)\src\FluentValidation.Tests.dotnet" />
-    <exec Command="dotnet test --no-build" WorkingDirectory="$(BaseDir)\src\FluentValidation.Tests.Mvc6.dotnet" />
+    <exec Command="dotnet test --no-build --configuration=$(Configuration)" WorkingDirectory="$(BaseDir)\src\FluentValidation.Tests.dotnet" />
+    <exec Command="dotnet test --no-build --configuration=$(Configuration)" WorkingDirectory="$(BaseDir)\src\FluentValidation.Tests.Mvc6.dotnet" />
+  </Target>
+  
+  <Target Name="TestMvcCore">
+    <exec Command="dotnet test --no-build --configuration=$(Configuration)" WorkingDirectory="$(BaseDir)\src\FluentValidation.Tests.Mvc6.dotnet" />
   </Target>
 
 </Project>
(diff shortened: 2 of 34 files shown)


batzen:
@JeremySkinner Any chance you could spare some time to have a look at this PR?

JeremySkinner:
I will try and do so soon...I don't get much time to work on FluentValidation and when I do, I've had to focus on aspnetcore compatibility.

batzen:
Didn't want to pressure you. Just wanted to make sure you don't forget about it. ;-)

JeremySkinner:
Thanks for doing this. I've had quick initial first look this morning. Some comments:

- In MessageFormatter.cs, the switch statement in GetPlaceholder could use the PropertyName/PropertyValue constants rather than duplicating the names.
https://github.com/JeremySkinner/FluentValidation/pull/334/commits/746fc512c78196054a9ebde2628f1a7688ebfb6f

- In PropertyValidatorContext.cs, is there a benefit of making the MessageFormatter lazily instantiated? The MessageFormatter is used by pretty much all of the validators, so would this one make more sense to stay as it was before?
https://github.com/JeremySkinner/FluentValidation/pull/334/commits/e52d12663cd2f5fd772e53df551c404ea264fdc3

- In PropertyRule.cs, the caching of the DisplayName is going to cause problems as this means you can no longer lazily load the property name from a custom IStringSource. I've seen people use custom implementations that aren't necessarily tied to culture, so this change will need removing. 

- In PropertyValidator.cs, moving the AppendPropertyName and AppendArgument("PropertyValue") calls into CreateValidationError is a subtle but large breaking change, as anyone overriding CreateValidationError will no longer have these values automatically available, so this will need to go back to how it was before.

batzen:
- Will use those constants.

- Instantiating it lazily has much benefit. MessageFormatter is only needed if you create a ValidationFailure and/or want to format a message. Thus it's not needed for validations that succeed.

- People can still use their own IStringSource. But their result would be cached. The only problem that could pop up is if their IStringSource returns different results when called at a different point in time. My guess is that there are people using IStringSource but not returning different results at different points in time. I could add an option to ValidatorOptions called "CacheDisplayNameInPropertyRule" if you'd prefer that. Caching DisplayName contributes much to these performance improvements. Maybe you got a better idea on how to detect if the cache is valid or not.

- But you agree that AppendPropertyName etc. should only be called if validation does not succeed, don't you? If you agree i would add the calls back to the validate methods, introduce a new virtual method called "PrepareMessageFormatterForValidationError" and call that method conditionally.

JeremySkinner:
> Instantiating it lazily has much benefit. MessageFormatter is only needed if you create a ValidationFailure and/or want to format a message. Thus it's not needed for validations that succeed.

Ok great, I'm happy with that. 

> My guess is that there are people using IStringSource but not returning different results at different points in time

There are also people who are returning different results at different points in time....that's the whole point of having IStringSource. I don't want this one cached, and I'd rather not have an extra configuration entry for it, so we'll leave it as per the original implementation.

> But you agree that AppendPropertyName etc. should only be called if validation does not succeed, don't you? 

Yes, that's sounds good.

JeremySkinner:
How about I add an interceptor that allows you to wrap IStringSource implementations with your own version that provide caching? This way you can add the caching into your project without it affecting the core implementation?

batzen:
I get your point and will revert my DisplayName caching changes.

There is no need to add a way of interception as that already exists. ;-)
Replacing DisplayNameResolver with a CachedDisplayNameResolver or something like that already works.
Writing that one was my first try while improving performance in that area but it didn't work very well.
Maybe i'll give it another try as soon as i got some time.

batzen:
Made the changes you requested.

Could you @JeremySkinner give it another look?

JeremySkinner:
Looks good, thanks. Could you squash these into a single commit? I'll then get this merged in.

batzen:
Damn, something went wrong while squashing...

batzen:
Should be ok now.

JeremySkinner:
Merged, thanks!

Issue #339, opened 2016-10-18
Mvc Core Integration skip abstract or open generic validators


gideonkorir:
First of all thanks for the awesome work. I'm getting an **ArgumentException** when I add a generic or abstract validator in my project.; the code fails at `IMvcBuilder.AddFluentValidation` because FluentValidation is trying to add the abstract/open generic validator to ServiceCollection 
![fluent_exception](https://cloud.githubusercontent.com/assets/1640535/19471530/4f735e26-952c-11e6-97f8-afcbbecf5273.PNG)

gideonkorir:
We are not filtering out abstract or open generic types in `FluentValidationMvcExtensions.RegisterTypes` I'm happy to put in a pull request if that's ok

JeremySkinner:
Yes, that would be really helpful if you could- thanks. I thought I'd
already filtered them out but obviously not!

Thanks

geirsagberg:
Looks like this could be closed?

JeremySkinner:
Correct

PR #340, opened 2016-10-18
Filtered out abstract or open generic types #339

gideonkorir:
- Filtered out abstract and generic type definitions


JeremySkinner:
Applied, thanks!

Issue #81, opened 2015-06-13
MustAsync makes browser hanging on without resolving


mdmoura:
Hello,

I have the following Validator on a MVC5 site:

```
public class CatalogCreateModelValidator : AbstractValidator<CatalogCreateModel> {

  public CatalogCreateModelValidator(Func<IMediator> mediator) {      
    RuleFor(x => x.Name)
      .NotEmpty().WithMessage("The name is required")
      .MustAsync(name => mediator().SendAsync(new CatalogNameValidationModel(name)))
      .WithMessage("The name is unavailable");
  }
}
```

When the name is empty I get an error on my view.
When name is not empty the browser hangs on "loading" without showing the error.

It seems it hangs on MustAsync ... The validation of the CatalogNameValidationModel is:

```
public class CatalogNameValidationModelHandler : IAsyncOrderHandler<CatalogNameValidationModel, Boolean> {

  private readonly IDbContext _context;

  public CatalogNameValidationModelHandler(IDbContext context) {
    _context = context;
  }

public async Task<Boolean> Handle(CatalogNameValidationModel model) {
   return !await _context.Catalogs.AnyAsync(x => x.Name == model.Name);
} // Handle
```

  }

My application works fine with Async in controllers ...

The only think it is not working is the MustAsync validation.

Thank You,
Miguel

JeremySkinner:
Could you put together a unit test that reproduces the problem? I'll then be able to look at it further.

mdmoura:
I just created a simple project based on ASP.NET MVC template project where I was able to replicate the error: https://github.com/mdmoura/FluentValidation.Async

The validator is the following:

  public class HomeIndexModelValidator : AbstractValidator<HomeIndexModel> {
    public HomeIndexModelValidator() {
      RuleFor(x => x.Name)
        .NotEmpty().WithMessage("The name is required")
        .MustAsync(x => new Task<Boolean>(() => { Task.Delay(1000); return true; })).WithMessage("Must Async validation is false");
    }
  }

If you start the project you will get a form.
If you submitted it without filling with will get the message "The name is required".
If you fill the name then the browser hangs out when, i think, must async is being called.

Any idea why this happens?

JeremySkinner:
Thanks - I'll take a look at this and try to get back to you in the next couple of weeks.

mdmoura:
That would be great! I have two applications using Async which break with FV's MustAsync.

Thank You

JeremySkinner:
[...]

PR #343, opened 2016-10-19
Added support for a custom async predicate validator

maca88:
The reason behind this feature is that the current AsyncPredicateValidator can cause a deadlock when we define a MustAsync that is validated using the Validate method. The deadlock will occur only in a single threaded context (ie. Windows UI thread or ASP.NET request context). There is also an issue #81 opened for this problem. 
So with this feature we are able to define a custom async predicate validator that can avoid a deadlock. [HERE](https://gist.github.com/maca88/4f10ec3ac66dd313cef52f4556b0fdd1) is an example by using Nito.AsyncEx package.

### DIFF
--- a/src/FluentValidation/DefaultValidatorExtensions.cs
+++ b/src/FluentValidation/DefaultValidatorExtensions.cs
@@ -376,7 +376,7 @@ namespace FluentValidation {
 		/// <returns></returns>
 		public static IRuleBuilderOptions<T, TProperty> MustAsync<T, TProperty>(this IRuleBuilder<T, TProperty> ruleBuilder, Func<T, TProperty, PropertyValidatorContext, CancellationToken, Task<bool>> predicate) {
 			predicate.Guard("Cannot pass a null predicate to Must.");
-			return ruleBuilder.SetValidator(new AsyncPredicateValidator((instance, property, propertyValidatorContext, cancel) => predicate((T) instance, (TProperty) property, propertyValidatorContext, cancel)));
+			return ruleBuilder.SetValidator(ValidatorOptions.AsyncValidatorFactory((instance, property, propertyValidatorContext, cancel) => predicate((T) instance, (TProperty) property, propertyValidatorContext, cancel)));
 		}
 
 		/// <summary>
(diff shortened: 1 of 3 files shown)


JeremySkinner:
Thanks for this. Before I make this change, could you confirm if this pull request solves your issue? 
https://github.com/JeremySkinner/FluentValidation/pull/344

Thanks

maca88:
Actually the other pull request does not solve the problem entirely and also does not compile on .NET 4, as Task.Run was added in .NET 4.5. Task.Run will execute the async validation in a new thread and prior to .NET 4.6 the new thread won't share the same culture as it is bounded to the thread. What I wanted is that the async validation is executed in the same thread without deadlocks, but to achieve this is not quite simple, that why I leaved the default one as it is.

JeremySkinner:
Thanks for the reply...we don't actually support .NET older than 4.5 anymore so not worried about .net 4.0 compatibility. Do you have an example of the code that fixes this properly? I'd rather have the fix included (even if it's non trivial) than introducing the delegate.

maca88:
[HERE](https://blogs.msdn.microsoft.com/pfxteam/2012/01/20/await-synchronizationcontext-and-console-apps/) is an example, but is not a portable solution as for example BlockingCollection does not exist in .NET 4.5 portable. For a portable solution you can check [AsyncEx](https://github.com/StephenCleary/AsyncEx/blob/master/Source/Nito.AsyncEx%20(NET45%2C%20Win8%2C%20WP8%2C%20WPA81)/AsyncContext.cs) library, but I don't think that is worth coping the whole thing.

maca88:
If the named delegate is a problem I can switch back to Func<> delegate in order to avoid a breaking change.

JeremySkinner:
It's not really the name of the delegate that's the problem, but it seems unnecessary. 

I'd like to have the "default" solution be something that covers most use cases (I thought the other pull request would be this, but I'm open to alternatives). 

For the case where you want to do something custom, I'd suggest not having a separate method on ValidatorOptions that does this, but instead call RuleFor(x => x.Foo).SetValidator(...) directly passing in your custom validator (MustAsync is just a friendly wrapper around SetValidator anyway). This is the usual way that FV works with extensibility...have fairly sensible defaults, but then it allows you to drop down to the underlying API when you need to.

maca88:
I will close this PR as I created another one, that fix the default AsyncPredicateValidator

PR #344, opened 2016-10-19
Using Task.Run in AsyncValidatorBase to avoid deadlock

abatishchev:
As I understand calling just ``Task.Result`` is unsafe and might cause a deadlock. Wrapping it with `Task.Run(() => ..).Result` is safe.


JeremySkinner:
Thanks!

JeremySkinner:
So this change has actually been superceded by this one: https://github.com/JeremySkinner/FluentValidation/pull/347

Could you check this out and confirm that it works OK for you?

PR #347, opened 2016-10-20
Support sync running async validations using one thread

maca88:
Implementation taken from [HERE](http://stackoverflow.com/questions/5095183/how-would-i-run-an-async-taskt-method-synchronously/5097066#5097066). I did some minor modifications to preserve the original stacktrace by using ExceptionDispatchInfo.Capture.

### DIFF
--- a/src/FluentValidation.NetStandard/FluentValidation.NetStandard.csproj
+++ b/src/FluentValidation.NetStandard/FluentValidation.NetStandard.csproj
@@ -101,6 +101,9 @@
     <Compile Include="..\FluentValidation\Resources\StaticStringSource.cs">
       <Link>StaticStringSource.cs</Link>
     </Compile>
+    <Compile Include="..\FluentValidation\TaskHelpers\AsyncHelpers.cs">
+      <Link>AsyncHelpers.cs</Link>
+    </Compile>
     <Compile Include="..\FluentValidation\TaskHelpers\TaskHelpers.cs">
       <Link>TaskHelpers.cs</Link>
     </Compile>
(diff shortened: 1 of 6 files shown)


JeremySkinner:
Looks good, thanks for this.

abatishchev:
I'm not that expert in async/await as [Stephen Cleary](http://stackoverflow.com/users/263693/stephen-cleary), he suggests to use `Task.Run().Result` and `Task.Run().Wait()`.

Also see [this answer](http://stackoverflow.com/a/17094801/41956) in the same thread on SO. Here's [another thread](http://stackoverflow.com/questions/22628087/calling-async-method-synchronously) which suggests the same.

In general, messing up with scheduler manually looks suspicious and unnecessary to me.

maca88:
> If you do find yourself in this corner, I would recommend using something like Dispatcher.PushFrame for WPF apps, looping with Application.DoEvents for WinForm apps, and for the general case, my own AsyncContext.Run.
> 

In the end Stephen suggested his own AsyncContext.Run which is a nested message loop.
I tested this code in a WPF and ASP.NET application and haven't found any problem so far. So if anyone know an example where this code won't work please share it here.

maca88:
If you will revert this PR please al least change `.Result` to` .GetAwaiter().GetResult()` as it won't wrap any exception within an AggregateException, which complicates error handling.

abatishchev:
`GetAwaiter().GetResult()` is causing deadlocks badly, in two sister projects just last week. `Task.Run()` doesn't, it always worked fine from my experience.

maca88:
I meant to replace
`Task.Run(() => IsValidAsync(context, new CancellationToken())).Result`
with
`Task.Run(() => IsValidAsync(context, new CancellationToken())).GetAwaiter().GetResult()`

abatishchev:
Oh, right. Yes, this looks good to me definitely!

JeremySkinner:
This is in the latest beta on nuget (6.4.0-beta5)

PR #350, opened 2016-10-23
Added support for custom validation context

maca88:
The reason to have a custom validation context is to increase the validation performace when validating a collection of models for which we need to trigger a database query to do a check (ie. code existance in the db). With the current FV we cannot pass a custom data to a validation rule in order to avoid a query. Lets see an example:
```
public class ParentEntity {
	public List<ChildEntity> Children { get; set; } = new List<ChildEntity>();
}

public class ChildEntity {
	public string Code { get; set; }
}

public class ParentEntityValidator : AbstractValidator<ParentEntity> {
	public ParentEntityValidator(ISession dbSession) {
		RuleFor(o => o.Children).SetCollectionValidator(new ChildEntityValidator(dbSession));
	}
}

public class ChildEntityValidator : AbstractValidator<ChildEntity> {
	public ChildEntityValidator(ISession dbSession) {
		RuleFor(m => m.Code)
			.Must((entity, code, ctx) => {
				return dbSession.Query<ChildEntity>().Any(o => o.Code == code);
			})
			.WithMessage("Code does not exist in the database");
	}
}
```
In the above example we need to trigger a query for each child which is not scalable. This validation could be improved by fetching the children codes from the database with one query and store the result in a custom context:
```
static void Main(string[] args) {
	...
	// Create entities to validate 
	var parent = new ParentEntity();
	for (var i = 0; i < 500; i++) {
		parent.Children.Add(new ChildEntity {Code = $"Code{i}"});

	}
	
	// Get all children codes that we will use in the query
	var childCodes = parent.Children.Select(o => o.Code).ToList();
	// Execute a query that will find all existing codes from the database
	var existingCodes = new HashSet<string>(dbSession.Query<ChildEntity>().Where(o => childCodes.Contains(o)));

	// Create a custom validation context with the fetched existing codes
	var rootCtx = new RootValidationContext<ParentEntity, CachedData>(parent, new CachedData(existingCodes));

	// Validate the parent
	var result = new ParentEntityValidator().Validate(rootCtx);
}

// New child validator
public class ChildEntityValidator : AbstractValidator<ChildEntity> {
	public ChildEntityValidator() {
		RuleFor(m => m.Code)
			.Must((entity, code, ctx) => {
				// Get the cached existing codes in order to avoid a db query
				var cctx = ctx.ParentContext as ICustomValidationContext;
				var rootCtx = cctx.GetRootData<CachedData>();
				if (!rootCtx.ExistingCodes.Contains(code)) {
					return false;
				}
				return true;
			})
			.WithMessage("Code does not exist in the database");
	}
}
```
With this approach the validation will be much faster as we will trigger the database only once. Also the parent and child validators can now be registered as singleton, as the new validators don't depend on a database connection anymore. Having singleton validators will additionally speedup the validaton process.
[HERE](https://gist.github.com/maca88/9e4551da22cf53056143bef76148a452) you can find the whole example.
I know that there are a lot of breaking changes, but IMO this feature is a must for complex business applications where we have a lot of validation rules like this.

### DIFF
--- a/src/FluentValidation.Mvc4/FluentValidationModelValidator.cs
+++ b/src/FluentValidation.Mvc4/FluentValidationModelValidator.cs
@@ -25,7 +25,7 @@ namespace FluentValidation.Mvc {
 			if (Metadata.Model != null) {
 				var selector = customizations.ToValidatorSelector();
 				var interceptor = customizations.GetInterceptor() ?? (validator as IValidatorInterceptor);
-				var context = new ValidationContext(Metadata.Model, new PropertyChain(), selector);
+				var context = ValidatorOptions.ValidationContextFactory.Get(Metadata.Model, new PropertyChain(), selector);
 
 				if(interceptor != null) {
 					// Allow the user to provide a customized context
--- a/src/FluentValidation.Mvc4/IValidatorInterceptor.cs
+++ b/src/FluentValidation.Mvc4/IValidatorInterceptor.cs
@@ -13,7 +13,7 @@ namespace FluentValidation.Mvc {
 		/// <param name="controllerContext">Controller Context</param>
 		/// <param name="validationContext">Validation Context</param>
 		/// <returns>Validation Context</returns>
-		ValidationContext BeforeMvcValidation(ControllerContext controllerContext, ValidationContext validationContext);
+		IValidationContext BeforeMvcValidation(ControllerContext controllerContext, IValidationContext validationContext);
 
 		/// <summary>
 		/// Invoked after MVC validation takes place which allows the result to be customized.
@@ -23,6 +23,6 @@ namespace FluentValidation.Mvc {
 		/// <param name="validationContext">Validation Context</param>
 		/// <param name="result">The result of validation.</param>
 		/// <returns>Validation Context</returns>
-		ValidationResult AfterMvcValidation(ControllerContext controllerContext, ValidationContext validationContext, ValidationResult result);
+		ValidationResult AfterMvcValidation(ControllerContext controllerContext, IValidationContext validationContext, ValidationResult result);
 	}
 }
--- a/src/FluentValidation.Mvc4/PropertyValidatorAdapters/FluentValidationPropertyValidator.cs
+++ b/src/FluentValidation.Mvc4/PropertyValidatorAdapters/FluentValidationPropertyValidator.cs
@@ -41,7 +41,7 @@ namespace FluentValidation.Mvc {
 					DisplayName = Rule == null ? null : Rule.DisplayName,
 				};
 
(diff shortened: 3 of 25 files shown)


JeremySkinner:
Thanks for the suggestion. 

I'm really not keen on introducing such a large breaking change...particularly introducing interface contracts for ValidationContext is something I absolutely want to avoid as it makes adding any new properties a breaking change in the future. But I'm wondering if the same result could be achieved with a simpler solution. 

If I were to add a dictionary property to ValidationContext, this could be used to store any kind of custom state (including the database results from your example) provided that the clone mechanism etc takes account of this. Would this work for you?

maca88:
Yes, a dictionary would be good enough for this purpose.

JeremySkinner:
I've pushed a new beta build to nuget that has the dictionary property in place (RootContextData). Please give it a try and let me know if you run into any issues.

maca88:
Thanks for your fast response, very appreciated!
Unfortunately the dictionary is not copied for every case as there are a lot of examples where the dictionary is not copied (ie. [HERE](https://github.com/JeremySkinner/FluentValidation/blob/master/src/FluentValidation/Internal/DelegateValidator.cs#L113) or [HERE](https://github.com/JeremySkinner/FluentValidation/blob/master/src/FluentValidation/AbstractValidator.cs#L73))
I've created a simple [console application](https://gist.github.com/maca88/20c4903965810c025d8ae3a2e7a5633a) to show where it works and where it doesn't.
In addition to that, it would be great if we had an universal interceptor when the root model is being validated like we have it for the [MVC](https://github.com/JeremySkinner/FluentValidation/blob/0486d739df1dc212e70800eb83903339e14d43d8/src/FluentValidation.Mvc3/FluentValidationModelValidator.cs#L33) in order to set our custom data. For example when using WebApi we cannot pass a custom data as [the validation](https://github.com/JeremySkinner/FluentValidation/blob/0486d739df1dc212e70800eb83903339e14d43d8/src/FluentValidation.WebApi/FluentValidationModelValidator.cs#L42) is done automatically.

JeremySkinner:
Ok that's fixed, please give it another try.

maca88:
Its working now, thanks.

JeremySkinner:
Thanks for the suggestion.

I'm really not keen on introducing such a large breaking
change...particularly introducing interface contracts for ValidationContext
is something I absolutely want to avoid as it makes adding any new
properties a breaking change in the future. But I'm wondering if the same
result could be achieved with a simpler solution.

If I were to add a dictionary property to ValidationContext, this could be
used to store any kind of custom state (including the database results from
your example) provided that the clone mechanism etc takes account of this.
Would this work for you?

PR #366, opened 2016-12-07
Fix Key of AccessorCache for complex expressions x => x.A.B.Foo

sjancke:
When using RuleFor(x => x.A.B.Foo) and RuleFor(x => x.A.C.Foo),
the AccessorCache does not differentiate between both expressions and
returns a wrong one (the first) for x => x.A.C.Foo.


sjancke:
Fix for performance improvements merged within issue #331

JeremySkinner:
Thanks!

PR #367, opened 2016-12-07
IBAN account number validator

edgars-pivovarenoks:
Decided to bring International Bank Account Number (IBAN) validator to FluentValidation. I think it would be usefull in many finance related projects especially in Europe.

Credit goes to following CodeProject post :
http://www.codeproject.com/Tips/775696/IBAN-Validator

Summary of changes :
- Added IbanValidator class,
- Added IsIBAN() extension method,
- Unit tests with many valid IBAN accouns.

Will be glad to hear feedback and suggestions.

PS. What is current consensus about built-in validators? Keep it to the minimum or adding common uses cases is welcomed?

### DIFF
--- a/src/FluentValidation.Tests/IbanValidatorTests.cs
+++ b/src/FluentValidation.Tests/IbanValidatorTests.cs
@@ -1,0 +1,94 @@
+namespace FluentValidation.Tests
+{
+    using System.Globalization;
+    using System.Linq;
+    using System.Threading;
+    using Xunit;
+
+
+    public class IbanValidatorTests
+    {
+        TestValidator validator;
+
+        public IbanValidatorTests()
+        {
+            CultureScope.SetDefaultCulture();
+
+            validator = new TestValidator {
+                v => v.RuleFor(x => x.IbanAccount).IsIBAN()
+            };
+        }
+
+        [Fact]
+        public void IsValidTests()
+        {
+            // Empty values
+            validator.Validate(new Person { IbanAccount = null }).IsValid.ShouldBeTrue(); // Optional values are always valid
+            validator.Validate(new Person { IbanAccount = string.Empty }).IsValid.ShouldBeTrue(); // empty string
+            validator.Validate(new Person { IbanAccount = "\n\r" }).IsValid.ShouldBeTrue(); // new line
+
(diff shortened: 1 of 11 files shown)


src/FluentValidation.Tests/IbanValidatorTests.cs
```diff
@@ -0,0 +1,94 @@
...
+            };
+        }
+
+        [Fact]
+        public void IsValidTests()
+        {
+            // Empty values
+            validator.Validate(new Person { IbanAccount = null }).IsValid.ShouldBeTrue(); // Optional values are always valid
+            validator.Validate(new Person { IbanAccount = string.Empty }).IsValid.ShouldBeTrue(); // empty string
+            validator.Validate(new Person { IbanAccount = "\n\r" }).IsValid.ShouldBeTrue(); // new line
```
wizofaus: What's the use case for empty string and \n\r (which is normally CRLF = \r\n) being considered valid?

src/FluentValidation.Tests/IbanValidatorTests.cs
```diff
@@ -0,0 +1,94 @@
...
+
+            // Valid sample IBAN account numbers
```
wizofaus: Needless to say I hope none of these are real account numbers!

src/FluentValidation/Validators/IbanValidator.cs
```diff
@@ -0,0 +1,76 @@
...
+
+            if(bankAccount.Length < 5)
+                return false;
+
+            else if (System.Text.RegularExpressions.Regex.IsMatch(bankAccount, "^[a-zA-Z0-9]*$"))
+            {
+                bankAccount = bankAccount.Replace(" ", string.Empty);
+                string bank =
+                bankAccount.Substring(4, bankAccount.Length - 4) + bankAccount.Substring(0, 4);
+                int asciiShift = 55;
```
wizofaus: I thought IBANs had to start with a set of predefined two-letter country codes?

JeremySkinner:
Cleanup old pull request; closing as this isn't considered a core validator should be distributed as part of the standard FV package.

PR #370, opened 2016-12-12
Add test and make it pass through implementation of evaluating Depend…

### LINKED ISSUE #253: Dependent rules don't run when validating asynchronously
``` c#
        [TestMethod]
        public async Task TestAsyncWithDependentRules()
        {
            validator.RuleFor(o => o)
                .NotNull()
                .DependentRules(d =>
                {
                    d.RuleFor(o => o.Address).NotNull();
                });

            var person = new Person();

            var result = await validator.ValidateAsync(person);

            Assert.IsFalse(result.IsValid);
        }
```

Dashue:
Add test and make it pass through implementation of evaluating DependentRules.

Would something like this take care of the dependent rules when async issue? Trying to resolve #253

### DIFF
--- a/src/FluentValidation.Tests/RuleBuilderTests.cs
+++ b/src/FluentValidation.Tests/RuleBuilderTests.cs
@@ -184,6 +184,49 @@ namespace FluentValidation.Tests {
 			validator.Verify(x => x.ValidateAsync(It.Is<PropertyValidatorContext>(c => (string)c.PropertyValue == "Foo"), It.IsAny<CancellationToken>()));
 
 		}
+
+		[Fact]
+		public async Task TestAsyncWithDependentRules_SyncEntry()
+		{
+			var validator = new TestValidator();
+			validator.RuleFor(o => o)
+				.NotNull()
+				.DependentRules(d => {
+					d.RuleFor(o => o.Address).NotNull();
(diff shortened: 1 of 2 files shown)


JeremySkinner:
I'm not sure that's going to work correctly...it only seems to be calling the dependent rules synchronously. As we're in the async path here, the dependentRules should be validated asynchronosly too.

JeremySkinner:
Also, that doesn't seem to behave consistently....the DependentRules should only be invoked if validation of the parent rule succeeds, but it seems to always run them regardless (assuming the async path) which I don't think is correct.

Dashue:
Cheers for the feedback!
I've updated the pr to what I think reflect what you were saying in regards to "should be validated asynchronosly too". I've also updated the test to break, which I believed to be caused by your second point with them being run all the time now. Should DependentRules only be run if all sync and async validators have passed?

Dashue:
I think that might do the trick! Might you throw an eye on it?

JeremySkinner:
Yes, I think they should only run if both the async and non-async validators all pass. 

I'll try and take a look at the updated PR later this week, thanks!

JeremySkinner:
Thanks, I've merged this in with 2 changes: 
- Excluded the tests from the Portable40 tests as they don't compile there
- Renamed RunDependentRules -> RunDependentRulesAsync

PR #426, opened 2017-03-20
Update AsyncValidatorBase.cs

rufusraja:


rufusraja:
The first letter of the first word in the identifier is lower case, while the first letter of every subsequent word is uppercase.

Issue #433, opened 2017-03-27
WebApi child object Modelstate not evaluated


danpowell88:
I'm trying to set up FV as metioned [here](https://levelnis.co.uk/blog/fluent-validation-with-autofac-basic-examples)

The validation appears to work ok for the top level objects but everything underneath never appears to be validated.

I know I could specify the SetValidator etc methods and new up the validator within but that seems to be against the whole point of using IOC.

From debugging I can see that WebApi never appears to ask the factory for a validator of the child types so it never botheres to validate them.

I've reproduced it in a project [here](https://github.com/danpowell88/fluentvalidationwebapiproblem)

See the [ValuesController post](https://github.com/danpowell88/fluentvalidationwebapiproblem/blob/master/WebApplication1/Controllers/ValuesController.cs#L30).

Sending 

```json
{
	Test: "a",
	Bar: { Pepsi : "j"}
}
```

Only triggers an error on "Test2" being null when it should also validate the Bar.Pepsi < 3

Any ideas?

JeremySkinner:
I'm pretty sure you'll need to use SetValidator for this (which is the recommended way of doing child-validation with FV)

Issue #365, opened 2016-11-30
Validation hierarchies in webapi


rich-zilla:
I have a hierarchy of models that im trying to validate as part of a web api request. Each model has its own validator. Is it possible to make fluent validation validate models that arent at the top of a hierarchy? 

The `IsValidatingProperty` method in `FluentValidationModelProvider` checks for the presence of a container and property name, meaning the validators for child models are never loaded.

morgen2009:
Is this what you are looking for?

```c#
using FluentValidation;
using FluentValidation.Results;

public class MasterDto {	
	public IList<DetailDto> Items;
}

public class DetailDto {
}

public class MasterDtoValidator : AbstractValidator<MasterDto>
{
	public MasterDtoValidator()
	{
		// write validation rules for master
		RuleFor(x => x.Items).SetCollectionValidator(x => new DetailDtoValidator());
	}
}

public class DetailDtoValidator : AbstractValidator<DetailDto>
{
	public DetailDtoValidator()
	{
		// write validation rules for detail
	}
}
```

PR #437, opened 2017-03-28
Fix child validators not being resolved in webapi

cj1x:
This addresses the issue in #433 and #365 where child validators are not resolved.

It seems that if `ValidatorFactory.GetValidator()` returns a validator, you would never want to short circuit validation for that model.


danpowell88:
I tried this against my issue and it appears to fix it, can we get it merged in?

JeremySkinner:
I don't think I'm going to have much time this week to work on FV, but I'll aim to get it done next week.

zihotki:
Are there any news about when it will be merged?

JeremySkinner:
I've merged it in - aiming to have it part of the 7.0 release.

PR #454, opened 2017-04-14
Tweak SplitPascalCase to handle UC sequences better

asherber:
I noticed that for property names like `URI`, `SplitPascalCase()` inserts lots of spaces: `U R I`. Changing the function to use a regex lets it be a little smarter about character groupings. Any of the following will be recognized as a group:
1. BOL followed by group of non-UC chars
1. UC char followed by group of non-UC chars 
1. One or more UC chars with lookahead UC char
1. One or more UC chars with lookahead EOL

I added some test cases to the relevant unit test to confirm that this works as expected.

### DIFF
--- a/src/FluentValidation.Tests/ExtensionTester.cs
+++ b/src/FluentValidation.Tests/ExtensionTester.cs
@@ -42,11 +42,16 @@ namespace FluentValidation.Tests {
 		public void Should_split_pascal_cased_member_name() {
 			var cases = new Dictionary<string, string> {
 				            {"DateOfBirth", "Date Of Birth"},
-				            {"DATEOFBIRTH", "D A T E O F B I R T H"},
+				            {"DATEOFBIRTH", "DATEOFBIRTH"},
 				            {"dateOfBirth", "date Of Birth"},
 				            {"dateofbirth", "dateofbirth"},
 							{"Date_Of_Birth", "Date_ Of_ Birth"},
-							{"Name2", "Name2"}
+							{"Name2", "Name2"},
+							{"ProductID", "Product ID"},
+							{"MyTVRemote", "My TV Remote"},
+							{"TVRemote", "TV Remote"},
+							{"XCopy", "X Copy"},
+							{"ThisXCopy", "This X Copy"},
 						};
 
 			foreach (var @case in cases) {
--- a/src/FluentValidation/Internal/Extensions.cs
+++ b/src/FluentValidation/Internal/Extensions.cs
@@ -23,6 +23,8 @@ namespace FluentValidation.Internal {
 	using System.Reflection;
 	using System.Text;
 	using System.Threading.Tasks;
+	using System.Text.RegularExpressions;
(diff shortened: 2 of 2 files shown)


JeremySkinner:
So this did actually used to be a regax but was changed for performance reasons.

JeremySkinner:
Sorry didn't mean to close this. 

I'll take a look at the performance implications of switching back to a regex again.

asherber:
I can also look at a version that walks through the string, like yours does. Looking ahead more than one character is easier in a regex, but I'll see what I can do.

JeremySkinner:
To be honest I'd be fine with going back to a regex but just need to do a few tests

asherber:
If you want to avoid the regex, I think the following does the trick:
```csharp
public static string SplitPascalCase(this string input)
{
    if (string.IsNullOrEmpty(input))
        return input;
    
    var retVal = new StringBuilder(input.Length + 5);

    for (int i = 0; i < input.Length; ++i)
    {
        var currentChar = input[i];
        if (char.IsUpper(currentChar))        
        {            
            if ((i > 1 && !char.IsUpper(input[i-1])) 
                    || (i + 1 < input.Length && !char.IsUpper(input[i+1])))
                retVal.Append(' ');
        }
        
        retVal.Append(currentChar);
    }

    return retVal.ToString().Trim();
}
```
IOW, if we get a UC and the _previous_ char was not UC, or if we get a UC and the _next_ char is not UC, put a space before it.

asherber:
And whoops ... I forgot to add two `using` directives in my original commit, because I'm working on two machines at once. Updated my branch to include them.

asherber:
It does look like the non-regex version is faster by a factor of 4 or 5, but I'm not sure how significant this is. I ran each function over 10,000 random 25-char strings; the regex version completed in about 95ms (over several runs), and the non-regex took 21ms. If I made the string only 15 chars, the regex version took about 63ms, and the non-regex took about 15ms.

In either case, we're looking at negligible differences for a more real-world use case of (e.g.) 100 15-char strings – 1ms or less for both methods. If you like the non-regex version better, I'll update my branch.

JeremySkinner:
Thanks. How much of a difference does it make if you use RegexOptions.Compiled?

asherber:
10,000x25 is about 78ms, 10,000x15 is about 54, so slightly quicker. Interestingly, 100x15 goes up a bit, to 2 ms. I guess the overhead of compiling the regex overwhelms the small run time. In a long-running app, though, you'd only take that hit once.

JeremySkinner:
Ok let's go with the non regex version for now. No need to update the pull request, I can just copy it in (I'm working in a different branch for the 7.0 release anyway)

asherber:
Okay, thanks!

asherber:
(You might want also the extra test cases I added.)

JeremySkinner:
Thanks, I've included those tests too. Changes are in the netstandard-migration branch and will be part of 7.0. Thanks again!

PR #455, opened 2017-04-14
Fix how the validator is resolved for classes that look like a collection

haefele:
In my case, I have a custom subclass of JObject, lets call it *Dog*, and a validator for it *DogValidator*.
But it would identify *Dog* as a collection (because JObject implements ICollection), and thus search for a validator for JToken.


JeremySkinner:
Thanks - I've merged this in.

PR #485, opened 2017-06-09
ShouldHaveChildValidator: Add any found child validators to the ValidationTestException

mattfrear:
I have made a minor change to the ShouldHaveChildValidator method. If any child validators are found, then the names of those are output to the ValidationTestException.


JeremySkinner:
Looks good, thanks

PR #497, opened 2017-06-27
Fix french translation for LessThanOrEqualValidator

Corbeau3000:
Simple translation fix to remove a word (sein : breast)


JeremySkinner:
Thanks! That's what happens when I completely trust the translations that people contribute.

PR #509, opened 2017-07-07
Added support for the Hindi language.

abhimjoshi:
Added support for the Hindi language for the default defined validators.


JeremySkinner:
Thanks!

PR #543, opened 2017-08-12
Code cleanup in FluentValidation.WebAPI

f1ana:
CreatePropertyModelName had unnecessary else if's/else and no need to null coalesce prefix after first check.


JeremySkinner:
Cleanup old pull request; no longer needed as WebApi integration has been rewritten to work with the new approach taken in more modern versionf of WebApi.

PR #601, opened 2017-11-03
when setting a conditional validator, register the type of validator …

Jestar342:
…instead of the property, for the 'validatorType' param.

Hi, when using something like `RuleFor(x => x.Foo).SetValidator(x => new FooValidator(x.Bar))` I noticed that `ShouldHaveChildValidator(x => x.Foo, typeof(FooValidator))` fails telling me `'FooValidator' expected, but found 'TypeOfFoo'` instead, so here is a fix.


JeremySkinner:
Thanks!

PR #602, opened 2017-11-03
Proposed fix for EnumValidator to handle negative values and other corner cases

wizofaus:
EnumValidator was previously crashing (OverflowException) if enum was defined with negative values, e.g.
[Flags]
enum Foo
{
 A= ~0, B = 3
}
It was also incorrectly considering 2 to be a valid Foo value (because it matched the bits in B).

New logic means that negative values will now work, but not values over Int64.MaxValue.


JeremySkinner:
Thanks!

Issue #600, opened 2017-11-03
When using other languges, some validators have incorrect localized messages (pt, fr)


FernandoNunes:
On the Localization of EN language (/Resources/Languages/EnglishLanguage.cs)
Translate<MinimumLengthValidator>("'{PropertyName}' must be more than {MinLength} characters. You entered {TotalLength} characters.");

Translate<MaximumLengthValidator>("'{PropertyName}' must be less than {MaxLength} characters. You entered {TotalLength} characters.");

On the Localization of ie. FR or PT language (Resources/Languages/PortugueseLanguage.cs)
Translate<MinimumLengthValidator>("'{PropertyName}' deve ter {MinLength} a {MaxLength} caracteres. Introduziu {TotalLength} caracteres.");
Translate<MaximumLengthValidator>("'{PropertyName}' deve ter {MinLength} a {MaxLength} caracteres. Introduziu {TotalLength} caracteres.");

When the message is displayed, the {MaxLength} property is not being filled and shows up: 
'Name' deve ter 0 a -1 caracteres. Introduziu 0 caracteres.

I don't know what is supposed to work, the Translation being fed with MaxLength attribute or if the Translation needs to be changed.

JeremySkinner:
The messages were wrong for some of the validators - I had previously corrected them in the English translation, but I'm not able to fix the translations for the other languages. If you're able to help with that, I'd appreciate it.

wizofaus:
I don't speak Portugese, but I'm fairly sure it should be "{PropertyName}' deve ser superior a {MinLength} caracteres. Introduziu {TotalLength} caracteres."  I'm looking at how to write a test to check that all the translations have the same substitutions as the English messages.

wizofaus:
Translation for language tr, key MaximumLengthValidator has parameters PropertyName,MinLength,MaxLength,TotalLength, expected PropertyName,MaxLength,TotalLength
Translation for language tr, key MinimumLengthValidator has parameters PropertyName,MinLength,MaxLength,TotalLength, expected PropertyName,MinLength,TotalLength
Translation for language sv, key NotEqualValidator has parameters PropertyName,CompairsonValue, expected PropertyName,ComparisonValue
Translation for language sv, key MaximumLengthValidator has parameters PropertyName,MinLength,MaxLength,TotalLength, expected PropertyName,MaxLength,TotalLength
[...]

PR #603, opened 2017-11-04
Partial fix for #600 - incorrect localized messages

wizofaus:
Need translators to provide proper messages for MinimumLengthValidator & MaximumLengthValidator for most languages.

### DIFF
--- a/src/FluentValidation.Tests/LanguageManagerTests.cs
+++ b/src/FluentValidation.Tests/LanguageManagerTests.cs
@@ -101,6 +101,32 @@
 		}
 
 
+		[Fact]
+		public void All_localizatons_have_same_parameters_as_English() {
+
+			LanguageManager manager = (LanguageManager)_languages;
+			var languages = manager.GetSupportedLanguages();
+			var keys = manager.GetSupportedTranslationKeys();
+			
+			Assert.All(languages, l => Assert.All(keys, k => Check_Parameters_Match(l, k)));
+		}
+
+		void Check_Parameters_Match(string languageCode, string translationKey) {
+			string referenceMessage = _languages.GetString(translationKey);
+			string translatedMessage = _languages.GetString(translationKey, CultureInfo.CreateSpecificCulture(languageCode));
+			if (referenceMessage == translatedMessage) return;
+			var referenceParameters = ExtractTemplateParameters(referenceMessage);
+			var translatedParameters = ExtractTemplateParameters(translatedMessage);
+			Assert.False(referenceParameters.Count() != translatedParameters.Count() ||
+				referenceParameters.Except(translatedParameters).Any(),
+				$"Translation for language {languageCode}, key {translationKey} has parameters {string.Join(",", translatedParameters)}, expected {string.Join(",", referenceParameters)}");
+		}
+
+		IEnumerable<string> ExtractTemplateParameters(string message) {
+			message = message.Replace("{{", "").Replace("}}", "");
+			return message.Split('{').Skip(1).Select(s => s.Split('}').First());
+		}
+
 		public class CustomLanguageManager : LanguageManager {
 			public CustomLanguageManager() {
 				AddTranslation("en", "NotNullValidator", "foo");
--- a/src/FluentValidation.Tests/LanguageManagerTests.cs
+++ b/src/FluentValidation.Tests/LanguageManagerTests.cs
@@ -108,12 +108,12 @@
 			var languages = manager.GetSupportedLanguages();
 			var keys = manager.GetSupportedTranslationKeys();
 			
-			Assert.All(languages, l => Assert.All(keys, k => Check_Parameters_Match(l, k)));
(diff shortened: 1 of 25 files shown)


src/FluentValidation/Resources/Languages/GeorgianLanguage.cs
```diff
@@ -21,7 +21,7 @@ namespace FluentValidation.Resources {
 
 	internal class GeorgianLanguage : Language {
 
-		public override string Name => "ge";
+		public override string Name => "ka";
```
JeremySkinner: Good catch

src/FluentValidation.Tests/LanguageManagerTests.cs
```diff
@@ -101,6 +101,32 @@ public void Gets_translation_for_croatian_culture()
...
+		public void All_localizatons_have_same_parameters_as_English() {
+
+			LanguageManager manager = (LanguageManager)_languages;
+			var languages = manager.GetSupportedLanguages();
+			var keys = manager.GetSupportedTranslationKeys();
+			
+			Assert.All(languages, l => Assert.All(keys, k => Check_Parameters_Match(l, k)));
+		}
+
+		void Check_Parameters_Match(string languageCode, string translationKey) {
```
JeremySkinner: Rename to CheckParametersMatch

src/FluentValidation/Resources/Languages/ChineseSimplifiedLanguage.cs
```diff
@@ -27,12 +27,12 @@ internal class ChineseSimplifiedLanguage : Language {
...
 			Translate<LengthValidator>("'{PropertyName}' 的长度必须在 {MinLength} 到 {MaxLength} 字符，您已经输入了 {TotalLength} 字符。");
-			Translate<MinimumLengthValidator>("'{PropertyName}' 的长度必须在 {MinLength} 到 {MaxLength} 字符，您已经输入了 {TotalLength} 字符。");
-			Translate<MaximumLengthValidator>("'{PropertyName}' 的长度必须在 {MinLength} 到 {MaxLength} 字符，您已经输入了 {TotalLength} 字符。");
+			Translate<MinimumLengthValidator>("'{PropertyName}' 的长度必须在 {MinLength} 到 1000 字符，您已经输入了 {TotalLength} 字符。");
+			Translate<MaximumLengthValidator>("'{PropertyName}' 的长度必须在 0 到 {MaxLength} 字符，您已经输入了 {TotalLength} 字符。");
 			Translate<LessThanOrEqualValidator>("'{PropertyName}' 必须小于或等于 '{ComparisonValue}'。");
 			Translate<LessThanValidator>("'{PropertyName}' 必须小于 '{ComparisonValue}'。");
 			Translate<NotEmptyValidator>("请填写 '{PropertyName}'。");
-			Translate<NotEqualValidator>("'{PropertyName}' 不能和 '{PropertyValue}' 相等。");
+			Translate<NotEqualValidator>("'{PropertyName}' 不能和 '{ComparisonValue}' 相等。");
```
JeremySkinner: Good catch

src/FluentValidation.Tests/LanguageManagerTests.cs
```diff
@@ -101,6 +101,32 @@ public void Gets_translation_for_croatian_culture()
 		}
 
 
+		[Fact]
```
JeremySkinner: Should probably make this a [Fact(Skip="Explicit")] for now as we know it's going to fail until the messages are re-translated.

wizofaus: As of this PR they shouldn't fail though.  Are you seeing failures?
Sure, hard-coding 1000 isn't great but as it is it shows -1 which is surely worse...
I could hard-code an infinity symbol instead!
All those others issues with the translations were picked up by the test.
BTW there is a potential bug in my code if you happened to have a message that had 3 consecutive {'s - not sure if you're really worried about that possibility though.

wizofaus: Ready to push a new commit if you've made a call on the use 1000 (I did even try replacing with ∞ : not sure how, e.g.  "'Code' doit contenir entre 2 et ∞ caractères" would read to a French speaker though!)

JeremySkinner: @wizofaus Yes you were right, the test does pass (I'd modified something when I was testing).

Let's go with the 1000 characters

wizofaus:
Hmm, somehow a rogue commit ended up this branch, will have to re-do PR.

JeremySkinner:
I've merged this into the master branch with 1 changes.

The CheckParametersMatch method does not build in the FluentValidation.Tests.NetCoreapp1 test project as .net core 1.0 doesn't include the CultureInfo.CreateSpecificCulture method. I've modified the method to use the `new CultureInfo(languageCode)` constructor instead as this works on .net core 1, .net core 2 and regular framework.  

This will be in the 7.3 beta which I'll try and push to nuget shortly.

PR #604, opened 2017-11-06
593 mvc tests only

wizofaus:
Figured this was easier to do via PR after all once I got my head around sorting out my git branches

### DIFF
--- a/Build.cmd
+++ b/Build.cmd
@@ -1,7 +1,7 @@
 @echo off
 if "%1"=="PublishPackages" goto publish
 
-dotnet build .build\build.proj /p:Version=7.2 %*
+dotnet build .build\build.proj /p:Version=7.2.1 %*
 
 goto end
 
--- a/Changelog.txt
+++ b/Changelog.txt
@@ -1,3 +1,6 @@
+7.x - 
+Marked ValidationException as serializable when running on full framework
+
 7.2.1 - 24 October 2017
 Introduce MessageFormatterFactory 
 Additional language translations
--- a/Changelog.txt
+++ b/Changelog.txt
@@ -1,6 +1,7 @@
-7.2.1 - 
+7.2.1 - 24 October 2017
 Introduce MessageFormatterFactory 
 Additional language translations
+Additional checks to ensure HttpContext is registered
 
 7.2 - 5 October 2017 
 Updated AspNetCore integration to support IValidatableObject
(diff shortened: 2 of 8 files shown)


JeremySkinner:
Not sure what's happened here....the PR seems to include a few duplicate commits, and it also won't apply cleanly. I'll try and handle this manually later in the week (probably friday)

JeremySkinner:
I've merged in the test changes manually to the MvcValidation branch. I think the PR was based off the master branch which is why it won't merge cleanly.

JeremySkinner:
...although the json-based dictionary test fails because the keys don't match. 

The keys come back from the MVC model binding process as:
```
[1].Value.Id 
[1].Value.SomeBool 
```
but the test has the following assertions:

```
result.IsValidField("[456].Id").ShouldBeFalse();
result.IsValidField("[456].SomeBool").ShouldBeFalse();
```
...but those keys don't exist in the ModelState dictionary. I'm not sure what's correct, although I assume what's coming from ModelState is right as this is coming from MVC itself. What was the reason for expecting the other format?

wizofaus:
Yeah, that was more of "as a user this the sort of error I'd find useful" than anything.  If an error came back as, e.g.  "[1].Value.Name" should be less than 20 characters without telling me what the key was it wouldn't help much.  But sure, whatever makes the most sense for the way MVC model binding normally works.

JeremySkinner:
Ok thanks for clarifying. The modelstate keys are generated by the recursive binding process - that’s not something I can customise. I’ll update the test.

wizofaus:
Having said that, if you pass the dictionary straight to FormUrlEncodedContent or create a StringContent from the JSON serialization, then what gets transmitted is indistinguishable from an object with properties that have the names of the keys, e.g.
{"123":{"Id":1,"SomeBool":true},"456":{"Id":0,"SomeBool":false}}

Which then produces a dictionary with keys 123 and 456 and their expected associated values.
So there's some sense in considering the property names to be '123' and '456'.
But I would think sort of automatically-generated validation error message generated from this wouldn't be appropriate for displaying to general users (though it might suffice for a quick'n'dirty type app with a limited user base), so whatever's easiest to implement is a reasonable call to make.

JeremySkinner:
Just to clarify, we're talking here about the keys that are added to ModelState. These are generated internally by MVC's binding process - not by FluentValidation itself.

JeremySkinner:
The updated aspnetcore validation is part of 7.3-beta1.

PR #623, opened 2017-11-21
Czech Language translation

jeez27:
Thank you for your fast fix of bug we reported.
Here is corrected Czech language translation for you.


JeremySkinner:
Thank you! That's very much appreciated. I'll include that in the next release.

PR #635, opened 2017-11-30
Update SpanishLanguage.cs

fcomunoz:
The word "longitude" doesn't exist in spanish, the correct is "longitud". You can check the correct word in spanish: http://dle.rae.es/?id=NbKfdrQ

But, the appropiate word for the sentence is "largo", it means "long".


JeremySkinner:
Thanks!

PR #727, opened 2018-03-31
Add Ukrainian translation to default messages

medvedbeast:
Default messages translated to Ukrainian language


JeremySkinner:
Thanks!

Issue #732, opened 2018-04-12
IValidator.Validate with wrong implementation, ParamName is wrong in ArgumentNullException


souzara:
![image](https://user-images.githubusercontent.com/11986361/38676302-f4de46e2-3e30-11e8-9073-779ecaff70a7.png)

PR #733, opened 2018-04-12
#732 Issue - Fix

souzara:


src/FluentValidation/Internal/Extensions.cs
```diff
@@ -16,32 +16,40 @@
 // The latest version of this file can be found at https://github.com/jeremyskinner/FluentValidation
 #endregion
 
-namespace FluentValidation.Internal {
+namespace FluentValidation.Internal
+{
```
JeremySkinner: Please preserve the original formatting (braces on end of line).

src/FluentValidation/Internal/Extensions.cs
```diff
@@ -16,32 +16,40 @@
...
 
 	/// <summary>
 	/// Useful extensions
 	/// </summary>
-	public static class Extensions {
-		internal static void Guard(this object obj, string message) {
-			if (obj == null) {
-				throw new ArgumentNullException(message);
+	public static class Extensions
+	{
```
JeremySkinner: Same as above - lease preserve the original formatting (braces on end of line) here and for the rest of this file.

souzara:
File Extensions.cs has been updated.
=)

PR #734, opened 2018-04-13
Should have child validator baseclass

janv8000:
For unit-testing purposes I want to construct my Validator with a mocked ChildValidator using eg. https://github.com/moq/moq4 .

Using ShouldHaveChildValidator gets me the following error:

> Expected property '{propertyName}' to have a child validator of type 'IValidator\`1.'. Instead found 'IValidator`1Proxy'
   at FluentValidation.TestHelper.ValidationTestExtension.ShouldHaveChildValidator

This PR should relax this check a little to allow setting an expectation on a base class but setting the validator to a derived class

### DIFF
--- a/src/FluentValidation/TestHelper/ValidatorTestExtensions.cs
+++ b/src/FluentValidation/TestHelper/ValidatorTestExtensions.cs
@@ -70,7 +70,7 @@ namespace FluentValidation.TestHelper {
 			var childValidatorTypes = matchingValidators.OfType<ChildValidatorAdaptor>().Select(x => x.ValidatorType);
 			childValidatorTypes = childValidatorTypes.Concat(matchingValidators.OfType<ChildCollectionValidatorAdaptor>().Select(x => x.ChildValidatorType));
 
-			if (childValidatorTypes.All(x => x != childValidatorType)) {
+			if (childValidatorTypes.All(x => !childValidatorType.GetTypeInfo().IsAssignableFrom(x.GetTypeInfo()))) {
 				var childValidatorNames = childValidatorTypes.Any() ? string.Join(", ", childValidatorTypes.Select(x => x.Name)) : "none";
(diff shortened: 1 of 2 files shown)


JeremySkinner:
Hi, thanks for the suggestion but this has broken the build. 

`GetTypeInfo` is not supported on all platforms - conditional compilation is needed to switch between GetType/GetTypeInfo depending on target platform. Please introduce conditional compilation to switch between the two, eg https://github.com/JeremySkinner/FluentValidation/blob/master/src/FluentValidation/AssemblyScanner.cs#L72-L88

If you can resolve this and commit an update that fixes the build then I'll merge this in.

src/FluentValidation/TestHelper/ValidatorTestExtensions.cs
```diff
@@ -70,7 +70,7 @@ public static class ValidationTestExtension {
 			var childValidatorTypes = matchingValidators.OfType<ChildValidatorAdaptor>().Select(x => x.ValidatorType);
 			childValidatorTypes = childValidatorTypes.Concat(matchingValidators.OfType<ChildCollectionValidatorAdaptor>().Select(x => x.ChildValidatorType));
 
-			if (childValidatorTypes.All(x => x != childValidatorType)) {
+			if (childValidatorTypes.All(x => !childValidatorType.GetTypeInfo().IsAssignableFrom(x.GetTypeInfo()))) {
```
JeremySkinner: Use of `GetTypeInfo` is not supported on all of FV's target platforms. Please introduce conditional compilation.

janv8000: I forgot to stage the extra `using` when committing. 
Do you want me to fixup the previous commit or leave as-is?

JeremySkinner:
That's great, thanks

Issue #590, opened 2017-10-23
Register open generics in ASP.NET Core


geirsagberg:
Given the following:

```csharp
public class Foo : IFoo {
  ...
}

public class FooValidator : AbstractValidator<IFoo> {
  ...
}
```

It would be nice to be able to resolve `IValidator<Foo>` from the ASP.NET Core DI Container.

I know the built-in DI container does not support open generics out of the box, so one would have to scan the assembly and explicitly register the validator for all concrete implementations of `IFoo`.

geirsagberg:
Seems like this would have to be implemented in the `AssemblyScanner`.

JeremySkinner:
The simplest thing is probably just to register these types with the container directly manually rather than relying on FV's built in AssemblyScanner

geirsagberg:
Yep, made an extension method:

```csharp
    public static class ServiceCollectionExtensions
    {
        public static void AddGenericValidator(this IServiceCollection services, Type genericType, Type validatorType, Assembly assembly)
        {
            var abstractValidatorType = typeof(IValidator<>);
            foreach (var implementationType in assembly.ExportedTypes.Where(type => !type.IsAbstract && !type.IsInterface && genericType.IsAssignableFrom(type)))
            {
                var genericValidatorType = abstractValidatorType.MakeGenericType(implementationType);
                services.AddTransient(genericValidatorType, validatorType);
            }
        }
    }
```

Always nice when this kind of stuff is built-in though; I know MediatR does some special handling to support open generics, maybe it can be reused?

JeremySkinner:
Yep, can definitely look at doing that for a future version

JeremySkinner:
@souzara yes if you could create a PR for me to review that'd be great.

JeremySkinner:
After reviewing this I think I won't include it in the core, as I feel it's not a common scenario and it's easy enough to register these manually if desired.

PR #736, opened 2018-04-17
Added ValidatorInjector, based in Issue #590 - Register open generics…

souzara:
@JeremySkinner as mentioned in Issue #590 , in this PR I don't register ValidatorInjector in IoC, like image below:
image
As you can see, it need be registered before to use, or users can instantiate by your self. What you think about that?

### DIFF
--- a/src/FluentValidation.Tests/ValidatorInjectorTest.cs
+++ b/src/FluentValidation.Tests/ValidatorInjectorTest.cs
@@ -1,0 +1,50 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace FluentValidation.Tests
+{
+    public class ValidatorInjectorTest
+    {
+		
+		[Fact]
+		public void Should_create_new_instance_of_ValidatorInjectorObjectValidator()
+		{
+			var validatorInjector = new ValidatorInjector<ValidatorInjectorObjectValidator>();
+			var model = new ValidatorInjectorObject
+			{
+				Id = 1,
+				Description = "Description"
+			};
+
+			Assert.NotNull(validatorInjector.Validator);
+
+			var validationResult = validatorInjector.Validator.Validate(model);
+			Assert.NotNull(validationResult);
+			Assert.True(validationResult.IsValid);
+		}
+
+	}
+	public interface IValidatorInjectorObject
+	{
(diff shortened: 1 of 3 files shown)


JeremySkinner:
Thanks, but this doesn't actually help with issue #590. 

The issue is specifically regarding extending the `AssemblyScanner` class to automatically register validators with open generic type definitions, which this doesn't do.

souzara:
Yes, realy it can't help with extending AssemblyScanner, but with a single a single line in ConfigureServices method in Startup class, like this: 
```charp
 services.AddSingleton(typeof(IValidatorInjector<>), typeof(ValidatorInjector<>));
```
Then inject it:
```charp
public anyConstructor(IValidatorInjecto<AnyValidator> validatorInjector)
{
      _validator = validatorInjector.Validator;
}
```
I thought that was clear in last message posted in #590, anyway can you suggest something else?

JeremySkinner:
Right, but that doesn't actually help. I think you may have misunderstood the problem that we're trying to solve. Let me try and explain.

At the moment, FluentValidation can automatically register all validators in your application with the container. This is done using the `AssemblyScanner` (which is wrapped by the `RegisterValidatorsFromAssemblyContaining` method). 

This automatic registration is useful because it means you don't have to register all your validators explicitly. So, instead of doing this:

```
services.AddMvc().AddFluentValidation();

services.AddTransient(typeof(IValidator<Foo>), typeof(FooValidator));
services.AddTransient(typeof(IValidator<Bar>), typeof(BarValidator));
services.AddTransient(typeof(IValidator<Baz>), typeof(BazValidator));
```

You can do this:

```
services.AddMvc().AddFluentValidation(fv => 
  fv.RegisterValidatorsFromAssemblyContaining<FooValidator>());
```

...and all the validators will be registered with the container. This means all your validators are wired up for the automatic integration with MVC, and you don't have to explicitly register each one. It also means you can inject any of your validators anywhere you need them. No need for an intermediary ValidatorInjector.

This all works correctly. 


Issue #590 is addressing the fact that this automatic registration doesn't handle *open generics*. That is, where the validator class has an unresolved generic type parameter. Eg:

```
public class FooValidator<T> : AbstractValidator<T> where T : IFoo { 

}
```

Here `FooValidator` is defined with an open generic type parameter. Validators like this won't be automatically registered using the above mechanism. This is what we're looking at supporting in a future release.

Hope that makes more sense.

PR #745, opened 2018-05-02
Added errors to the validation exception

luk355:
so that you can actually see why the validation failed.


JeremySkinner:
Thanks!

PR #749, opened 2018-05-09
Issue  #238

### LINKED ISSUE #238: TestHelper does not support RuleForEach
Hi @JeremySkinner  :wave: 

I've got a weird one here - [and here is a full repo of the issue](https://dotnetfiddle.net/DzW2nN).

When I try and use `ShouldHaveValidationErrorFor` in my unit test, and i'm validating against a collection using `RuleForEach` the normal validator _correctly_ captures the error (yay!) but the unit test extension method _fails_ to find that error message in the (internal) `Errors` collection. As such, it thinks there was no error for the property I'm testing.

The reason for this is that the Property name is `Foo` but because this is really a collection of `Foo`'s, the property name in the error is `Foo[0]` (or which ever item in the collection failed).

Here's a screen shot of the sample error:

```
RuleForEach(user => user.Links)
                .Must(LinkMustBeAUri) // given a string, return true/false if it's a URI.
                .WithMessage("Link '{PropertyValue}' must be a valid URI. eg: http://www.SomeWebSite.com.au");

...

_validator.ShouldHaveValidationErrorFor(l => l.Links, 
                    new [] { "!2134242", "adhgahsjdasdgjhad" }, 
                    ResidentialListingValidator.MinimumRuleSet);
```

`ValidationResultTester.cs` line: 46

image

So. to recap.
- I've got: `public string[] Links { get; set; }`
- I'm validating that the content of this collection is valid
- The validator _correctly_ finds the error. YAY!
- The test _extension_ method fails to figure this out. Damn :cry: 

As linked above - a full repo of the code is above.

AdemCatamak:
resolve #238

Bug occurs because of property name appears like propertyName[*]. Hence, I modied the property name (normalized it) before comparison.


AdemCatamak:
Hi @JeremySkinner ,

This PR has been open for a while. I wonder if this will be merged or not. Could you please inform me? If it is not going to be merged, what are the ways that I can improve it? If that's the case, could you kindly give me some advice on my improvements?

Thanks,
Best regards

JeremySkinner:
Hi, I’m afraid I haven’t had an opportunity to review this yet as I’ve been busy with other work. 

I work on FluentValidation on a voluntary basis in my spare time which means unfortunately I’m not always able to dedicate much time for it. I will get to this but it may not be for a while. Thanks for your patience.

AdemCatamak:
Thanks for your time and the voluntary work. Also, thanks for the fast response. Have a nice day.

JeremySkinner:
Merged, thanks!

PR #760, opened 2018-05-19
Validate and WhenAsync do not work properly together #759

### LINKED ISSUE #752: UnlessAsync and WhenAsync do not call for ShouldHaveValidationErrorFor
I have next code
```
RuleFor(x => x.Field).Must(x => x.Field == 0).WhenAsync((x, t) => service.ShouldInvokeAsync())
RuleFor(x => x.Field).Must(x => x.Field != 0).UnlessAsync((x, t) => service.ShouldInvokeAsync())
```

In case I am going to do assertion with `ShouldHaveValidationErrorFor` the statement the `WhenAsync` and `UnlessAsync` statements are not invoked.
So, it cause next issue: both rules are invoked when I perform ShouldHaveValidationErrorFor

Note: it works well in case I replace 
`WhenAsync((x, t) => service.ShouldInvokeAsync())` by `When(x => service.ShouldInvokeAsync().Result)`

### LINKED ISSUE #759: Validate and WhenAsync do not work properly together
Hi,
I examined the issue #752 and found another bug. If user uses `Validate` method which belongs to `AbstractValidator` and `WhenAsync` together, none of the Rule(s) that ‘WhenAsync’ method is used work. As a result, issue #752 occurs.

I would like to offer some modifications for `Validation` method in `PropertyRule.cs` file. One solution can be that validation method is checked. If it is async, `InvokePropertyValidatorAsync` method is called. If it is not async, `InvokePropertyValidator` method is called.
Another solution can be that, `Validation` method in `PropertyRule.cs` file can be changed such that: `return ValidationAsync(context, default(CancellationToken)).Result;`. This second solution works as well. One positive affect of this solution is that `Validation` method in `PropertyRule.cs` file is run from a single operation.

If you offer another solution, I would like to try to implement it.

AdemCatamak:
Fix #752 and Fix #759 

Check rule is async or sync and invoke validation rule according to the property in PropertyRuleValidator.


JeremySkinner:
Thanks, but this has broken the build. If you could fix the errors first then I’ll give this a review.

PR #772, opened 2018-06-01
WebAPI - Add ImplicitlyValidateChildProperties configuration

Emilien-M:
I need to validate only the root properties of my templates in my web API.

Currently, the default behavior validates all child properties.

So, I've added a configuration property in the provider that uses the current default behavior.


JeremySkinner:
Looks good, thanks (ignore the build failure - appveyor hasn't been updated to asp.net core 2.1 yet). 

I'm currently finishing the 7.6 release, and then I'll aim to get this in for 7.6.1. 

For future reference, it'd be great if you could stick to the project code style (braces end of lines).

Thanks again!

Emilien-M:
There is the code style for my changes and some others around.

JeremySkinner:
Applied, thanks!

PR #778, opened 2018-06-12
ToString method is implemented for ValidationResult class

AdemCatamak:
When the `ToString` method is called for an instance belonging to the` ValidationResult` class, error messages will be assembled using the `NewLine` separator.

If user want to give separator value, `string ToString (string separator)` method can be used.


JeremySkinner:
Merged, thanks.

Issue #790, opened 2018-06-17
Replace internal usage of Tuple with System.ValueTuple


JeremySkinner:
Requires taking a dependency on the System.ValueTuple package

PR #793, opened 2018-06-18
change to using ValueTuple

aaronincincy:
This is in response to #790.  I wasn't able to find a way to use the () syntax when pattern matching, so I had to use the ValueTuple<> generic directly.  If you know of a way around that I'd be glad to change it.


JeremySkinner:
Looks good, thanks @aaronincincy. I'll try and get this merged in this week.

JeremySkinner:
Merged, thanks!

PR #805, opened 2018-06-27
Tests does not pass when computer culture is not English.

AdemCatamak:
Localization leads validation result has different error message which is expected. I set culture info for 

ValidationResult gives a different error message than expected because of lacalization. Therefore, I set culture information for Test/App context.

### DIFF
--- a/src/FluentValidation.Tests.AspNetCore/StartupWithMvcValidationDisabled.cs
+++ b/src/FluentValidation.Tests.AspNetCore/StartupWithMvcValidationDisabled.cs
@@ -6,6 +6,8 @@ namespace FluentValidation.Tests.AspNetCore {
 	using Microsoft.Extensions.Logging;
 	using FluentValidation.AspNetCore;
 	using FluentValidation.Attributes;
+	using System.Globalization;
+	using Microsoft.AspNetCore.Localization;
 
 	public class StartupWithMvcValidationDisabled
     {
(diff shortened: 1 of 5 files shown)


src/FluentValidation.Tests.AspNetCore/StartupWithMvcValidationDisabled.cs
```diff
@@ -31,8 +33,15 @@ public void ConfigureServices(IServiceCollection services)
...
-            {
+	        CultureInfo cultureInfo = new CultureInfo("en-US");
+
+	        app.UseRequestLocalization(options => {
+		        options.DefaultRequestCulture = new RequestCulture(cultureInfo);
+		        options.SupportedCultures = new[] { cultureInfo };
+		        options.SupportedUICultures = new[] { cultureInfo };
+	        });
+
+			app.UseMvc(routes => {
```
AdemCatamak: In my computer, this line does not seen like this. I try to fix this

JeremySkinner: It's probably an issue tabs vs spaces. Turn on whitespace in VS and it should show you where the difference is (Edit -> Advanced -> View Whitespace)

AdemCatamak:
I believe that this time I will fix the style issues :)

JeremySkinner:
Merged, thanks.

AdemCatamak:
Hi Jeremy,

As you know, I encountered some style problems during development. The reason for some of these mistakes is that files have mixed values of tabs and spaces. (As you can see on the screen display)

If you do not see it as an unnecessary effort, I can translate all tabs values to spaces (or all spaces values to tabs).

JeremySkinner:
Thanks, they should use tabs for consistency with the rest of the project.

PR #826, opened 2018-07-03
Fix nested property selection

gnncl:
Validation fails if any nested property is null. This fix tries to handle that. Works on local test environment.

### DIFF
--- a/src/FluentValidation/Internal/Extensions.cs
+++ b/src/FluentValidation/Internal/Extensions.cs
@@ -151,36 +151,35 @@ namespace FluentValidation.Internal {
 
 #pragma warning disable 1591
 		public static Func<object, object> CoerceToNonGeneric<T, TProperty>(this Func<T, TProperty> func) {
-			return x => func((T)x);
+			return GetSafeValue<T, TProperty, object>(func);
 		}
 
 		public static Func<object, bool> CoerceToNonGeneric<T>(this Func<T, bool> func) {
-			return x => func((T)x);
+			return GetSafeValue<T, bool, bool>(func);
 		}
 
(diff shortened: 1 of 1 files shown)


src/FluentValidation/Internal/Extensions.cs
```diff
@@ -200,5 +199,27 @@ public static class Extensions {
 
 			return false;
 		}
+
+		private static Func<object, TResult> GetSafeValue<T, TProperty, TResult>(Func<T, TProperty> func) where TProperty : TResult {
+			return x => {
+				try {
+					return func((T)x);
+				}
+				catch (Exception) {
```
JeremySkinner: We shouldn't be catching general `Exception` here. Can you catch a more specific exception type?

gnncl:
Here is an example:

```
public class TestModel
{
    public NestedTestModel Nested { get; set; }
}

public class NestedTestModel
{
    public string Property { get; set; }
}

public class TestValidator : AbstractValidator<TestModel>
{
    public TestValidator()
    {
        RuleFor(x => x.Nested.Property)
            .NotEmpty();
    }
}

public class Test
{
    public void TestValidator()
    {
        var model = new TestModel();

        var result = validator.Validate(model);
    }
}
```
Before this change `validator.Validate(model);` throws `NullReferenceException`. After this fix it will validate the model and fill the result.

src/FluentValidation/Internal/Extensions.cs
gnncl: I have sent a new commit to catch only `NullReferenceException`.

JeremySkinner:
Closing as the current behaviour is correct. Please feel free to open an issue if you'd like me to clarify anything further.

PR #830, opened 2018-07-05
OnFailure extension method

Misiu:
This allows specifying action (function) that will be called when a specific rule fails.
In my case, I'm using it mostly for logging.

This PR is an initial version that contains only a single test, if more tests are needed please let me know.
Any improvements on how to simplify `OnFailureValidator` are more than welcome.


src/FluentValidation/Validators/OnFailureValidator.cs
```diff
@@ -0,0 +1,79 @@
...
+			_onFailureSimple = onFailure;
+		}
+
+		public OnFailureValidator(IPropertyValidator innerValidator, Action<T> onFailure)
+		{
+			_innerValidator = innerValidator;
+			_onFailureSuperSimple = onFailure;
+		}
+
+		public override IEnumerable<ValidationFailure> Validate(PropertyValidatorContext context)
```
JeremySkinner: As well as `Validate`, you'll also need to override `ValidateAsync` and add an appropriate asynchronous validation path that delegates to the inner validator's `ValidateAsync`. This will mean duplicating much of the code in here (but in an async manner)

src/FluentValidation/Validators/OnFailureValidator.cs
```diff
@@ -0,0 +1,79 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace FluentValidation.Validators
+{
+	public class OnFailureValidator<T, TProperty> : NoopPropertyValidator
+	{
+		private readonly IPropertyValidator _innerValidator;
+		private readonly Action<T> _onFailureSuperSimple;
```
JeremySkinner: Let's not have all these different delegate fields. Let's have just 1 that's an `Action<PropertyValidatorContext>`.  The others can all be supported using extension methods that provide delegates that wrap this one (various other methods in FV work in the same way)

src/FluentValidation/DefaultValidatorOptions.cs
```diff
@@ -352,6 +352,41 @@ public static class DefaultValidatorOptions {
...
+		/// <typeparam name="T"></typeparam>
+		/// <typeparam name="TProperty"></typeparam>
+		/// <param name="rule"></param>
+		/// <param name="onFailure"></param>
+		/// <returns></returns>
+		public static IRuleBuilderOptions<T, TProperty> OnFailure<T, TProperty>(this IRuleBuilderOptions<T, TProperty> rule, Action<T, PropertyValidatorContext> onFailure) {
+			return rule.Configure(config => config.ReplaceValidator(config.CurrentValidator, new OnFailureValidator<T, TProperty>(config.CurrentValidator, onFailure)));
+		}
+
+		public static IRuleBuilderOptions<T, TProperty> OnFailure<T, TProperty>(this IRuleBuilderOptions<T, TProperty> rule, Action<T, PropertyValidatorContext, string> onFailure, string errorMessage) {
```
JeremySkinner: Please add doc-comments to all public API fluent methods.

src/FluentValidation/DefaultValidatorOptions.cs
```diff
@@ -352,6 +352,41 @@ public static class DefaultValidatorOptions {
...
+			return rule.OnFailure(onFailure, errorMessage, null as object[]);
+		}
+
+		public static IRuleBuilderOptions<T, TProperty> OnFailure<T, TProperty>(this IRuleBuilderOptions<T, TProperty> rule, Action<T, PropertyValidatorContext, string> onFailure, string errorMessage, params object[] messageArgs) {
```
JeremySkinner: Remove this overload. Messages with custom args provided by `params object[] messageArgs)`  has been deprecated for quite some time. Instead, we would usually provide a callback that allows a user to make a call to `string.Format` if they want to. Eg:

```
OnFailure<T,TProperty>(this IRuleBuilderOptions<T, TProperty> rule, Action<T, PropertyValidatorContext, string> onFailure, Func<T, string> errorMessageBuilder) 
```

If the user wants to supply custom args for building the message, then they'd then specifify something like this for errorMessageBuilder:
```
x => string.Format("My Message", arg1, arg2);
```

This is how `WithMessage`, `WithLocalizedMessage` etc have worked since 7.x. 

The OnFailure validator would take a `Func<string>` parameter in its ctor instead of just a `string`.

src/FluentValidation/DefaultValidatorOptions.cs
```diff
@@ -352,6 +352,41 @@ public static class DefaultValidatorOptions {
...
+			var funcs = ConvertArrayOfObjectsToArrayOfDelegates<T>(messageArgs);
+			return rule.OnFailure(onFailure, errorMessage, funcs);
+		}
+
+		public static IRuleBuilderOptions<T, TProperty> OnFailure<T, TProperty>(this IRuleBuilderOptions<T, TProperty> rule, Action<T, PropertyValidatorContext, string> onFailure, string errorMessage, params Func<T, object>[] funcs) {
```
JeremySkinner: As above, remove this overload and make use of one that takes a `Func<T, string>` instead for error mesasges.

src/FluentValidation/DefaultValidatorOptions.cs
```diff
@@ -352,6 +352,41 @@ public static class DefaultValidatorOptions {
...
+			Func<T, TProperty, object>[] tmp = funcs.Select(func => new Func<T, TProperty, object>((instance, value) => func(instance))).ToArray();
+			return rule.OnFailure(onFailure, errorMessage, tmp);
+		}
+
+		public static IRuleBuilderOptions<T, TProperty> OnFailure<T, TProperty>(this IRuleBuilderOptions<T, TProperty> rule, Action<T, PropertyValidatorContext, string> onFailure, string errorMessage, params Func<T, TProperty, object>[] funcs) {
```
JeremySkinner: As above, remove this overload and make use of one that takes a `Func<T, string>` instead for error messages.

src/FluentValidation/DefaultValidatorOptions.cs
```diff
@@ -352,6 +352,41 @@ public static class DefaultValidatorOptions {
...
+		/// <summary>
+		/// Specifies custom method that will be called when specific rule fails
+		/// </summary>
+		/// <typeparam name="T"></typeparam>
+		/// <typeparam name="TProperty"></typeparam>
+		/// <param name="rule"></param>
+		/// <param name="onFailure"></param>
+		/// <returns></returns>
+		public static IRuleBuilderOptions<T, TProperty> OnFailure<T, TProperty>(this IRuleBuilderOptions<T, TProperty> rule, Action<T, PropertyValidatorContext> onFailure) {
+			return rule.Configure(config => config.ReplaceValidator(config.CurrentValidator, new OnFailureValidator<T, TProperty>(config.CurrentValidator, onFailure)));
```
JeremySkinner: Wrapping the validator using `ReplaceValidator` is fine, but it does mean that it breaks the TestHelper (specifically `ShouldHaveChildValidator`), so this PR will need to address that too.

src/FluentValidation.Tests/OnFailureTests.cs
```diff
@@ -0,0 +1,54 @@
...
+			_validator.RuleFor(person => person.Age).GreaterThanOrEqualTo(18).OnFailure((person, ctx, message) =>
+			{
+				Debug.WriteLine(message);
+				invoked += 1;
+			}, "You must be at least 18 years old. You entered: {0}", person=>person.Age);
+
+			_validator.Validate(new Person { Forename="John", Age = 17 });
+
+			invoked.ShouldEqual(4);
+		}
```
JeremySkinner: - [ ] Add tests for async workflow. 
- [ ] Add tests that ensure testhelper's `ShouldHaveChildValidator` still works

src/FluentValidation.Tests/OnFailureTests.cs
```diff
@@ -0,0 +1,54 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace FluentValidation.Tests
+{
+	using System.Diagnostics;
```
JeremySkinner: Can we make these `using` statements consistent? Some are inside the namespace, and some are outside. The rest of the project tends to have them all inside the namespace declaration.

Misiu:
@JeremySkinner thanks for the quick feedback.
In my case, I have a single OnFailure function that I call from multiple rules, so I needed that error message to pass a custom message to my function. Now when I think about it I could reuse message that is passed using `WithMessage` or default error message.

I've added async method, fixed code style, and removed unneeded overloads.

src/FluentValidation/Validators/OnFailureValidator.cs
```diff
@@ -0,0 +1,79 @@
+﻿using FluentValidation.Internal;
+using FluentValidation.Results;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace FluentValidation.Validators
+{
```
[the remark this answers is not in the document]
Misiu: Fixed

src/FluentValidation/Validators/OnFailureValidator.cs
Misiu: I've added async `Validate` method. Please take a look at it.

src/FluentValidation/DefaultValidatorOptions.cs
Misiu: Added

src/FluentValidation/DefaultValidatorOptions.cs
Misiu: This isn't used anymore - I've simplified usage.

src/FluentValidation.Tests/OnFailureTests.cs
Misiu: Fixed

src/FluentValidation.Tests/OnFailureTests.cs
Misiu: Async test added. I'll try adding test for second.

src/FluentValidation/Validators/OnFailureValidator.cs
Misiu: I need those 3 fields because I can pass 3 different actions:
`Action<T>`
`Action<T, PropertyValidatorContext>`
`Action<T, PropertyValidatorContext, string>`

How can I store those 3 types as one?

src/FluentValidation/DefaultValidatorOptions.cs
Misiu: Removed. I've simplified usage a bit.

src/FluentValidation/DefaultValidatorOptions.cs
Misiu: Can I achieve the same result without wrapping validator? If Yes then I can do that to not break anything else, if Not then what should I test for?

JeremySkinner: Without a wrapper you'd have to make changes to PropertyValidator which I wouldn't recommend. 

Stick with the wrapper, but basically you'd need to test that this works:

```
_validator.RuleFor(person => person.Address).SetValidator(new AddressValidator().OnFailure((person, ctx) => { });

// this should work:
_validator.ShouldHaveChildValidator(x => x.Address, typeof(AddressValidator));
// but it won't, because it'll find an instance of OnFailureValidator instead
```

And then the fix will be here: https://github.com/JeremySkinner/FluentValidation/blob/248682d7ef902eab48ccd5c9bb679e1d8026abd4/src/FluentValidation/TestHelper/ValidatorTestExtensions.cs#L74

That has a special case for unwrapping the validator if it's a ChildValidatorAdaptor. You'll need to add the OnFailureValidator here. It might actually be nice to add an interface for this, but I'm happy for you to hard code the type check for OnFailureValidator in there for now.

src/FluentValidation/Validators/OnFailureValidator.cs
JeremySkinner: Instead of defining a new delegate, you could use C# 7's support for local functions to make it a bit nicer.

Misiu: @JeremySkinner  something like this?

```
public class OnFailureValidator<T, TProperty> : NoopPropertyValidator {
	private readonly IPropertyValidator _innerValidator;
	private readonly Action<T, PropertyValidatorContext, string> _onFailure;

	public OnFailureValidator(IPropertyValidator innerValidator, Action<T, PropertyValidatorContext, string> onFailure) {
		_innerValidator = innerValidator;
		_onFailure = onFailure;
	}

	public OnFailureValidator(IPropertyValidator innerValidator, Action<T> onFailure)
		: this(innerValidator, (instance, ctx) => onFailure(instance)) {
	}

	public OnFailureValidator(IPropertyValidator innerValidator, Action<T, PropertyValidatorContext> onFailure)
		: this(innerValidator, (instance, context, errorMessage) => onFailure(instance, context)) {
	}

	public override IEnumerable<ValidationFailure> Validate(PropertyValidatorContext context) {
		var results = _innerValidator.Validate(context);
		if (!results.Any()) return results;
		var errorMessage = results.First().ErrorMessage;
		_onFailure((T)context.Instance, context, errorMessage);
		return results;
	}

	public override Task<IEnumerable<ValidationFailure>> ValidateAsync(PropertyValidatorContext context, CancellationToken cancellation) {
		var results = _innerValidator.ValidateAsync(context, cancellation);
		if (!results.Result.Any()) return results;
		var errorMessage = results.Result.First().ErrorMessage;
		_onFailure((T)context.Instance, context, errorMessage);
		return results;
	}
}
```

JeremySkinner: Yes but let's do it in the extension method rather than using constructor chaining for consistency with the rest of the framework. 

Also your async method is blocking. You shouldn't be using `results.Result` at any point. Use await instead.

Misiu: @JeremySkinner first part fixed :)
If I'll use await then ValidateAsync will have async modifier.

With all those changes my code is getting simpler and simpler :)

JeremySkinner:
Looks good, thanks! I'll have a quick play with it over the weekend and if all looks good I'll merge it in. Thanks again for contributing!

Misiu:
@JeremySkinner glad I could add something to already awesome library :)

Now the last part is `ShouldHaveChildValidator`.

I've tried replacing:
```
var childValidatorTypes = matchingValidators.OfType<ChildValidatorAdaptor>().Select(x => x.ValidatorType);
```
with
```
var childValidatorTypes = matchingValidators.Where(v=>v is ChildValidatorAdaptor || v is OnFailureValidator<T,TProperty>).Select(x => x.ValidatorType);
```
but I get error and I can't figure how to fix it.

JeremySkinner:
@Misiu That won't work because a) they're different types and b) OnFailureValidator doesn't expose a property called `ValidatorType` (which you're trying to access from the `select` statement), so the compiler won't know what to do with this

I'd suggest the following approach:

- Create an interface, something like `IExposesInnerValidator` with a readonly property called `ValidatorType` which returns `Type`.
- Implement this in `ChildValidatorAdaptor` (using the existing property)
- Implement this in `ChildCollectionValidatorAdaptor` (rename `ChildValidatorType` on this one to `ValidatorType` to match the interface)
- Implement this in `OnFailureValidator`:
  - The getter should turn check if `_innerValidator` is an `IExposesInnerValidatorType`. If so, return its `ValidatorType` property
  - If not, just return `_innerValidator.GetType()`
- Modify the TestHelper code to just check for `IExposesInnerValidator`, rather than having separate steps for each concrete class.

JeremySkinner:
Feel free to leave this last change. I needs it for something else I’m working on, so I’ll implement this last bit. Thanks again for doing this.

Misiu:
@JeremySkinner glad I could help :)
I've done one more commit that removes TProperty and cleans usings.

Just installed Resharper so I can apply some quick fixes if needed.

JeremySkinner:
I've rebased this with my latest changes and merged it in. I've also added compatibility with `ShouldHaveChildValidator` as we discussed: https://github.com/JeremySkinner/FluentValidation/commit/148c5aef356e0bda5a8a1bd76f5fbb529694b565

PR #835, opened 2018-07-11
Add Albanian Language to the default languages

h13xe:


h13xe:
Hello, sure thing. Here it is https://github.com/JeremySkinner/FluentValidation/pull/837

JeremySkinner:
Thanks, I've merged these in.

For future reference, it's better to add the extra commits to the same branch and pull request, rather than opening a new one, but I've combined them manually for now.

PR #844, opened 2018-07-23
Fix links to other docs` sections

Deilan:
Fix links from configuring page to built-in-validators section and localization section


JeremySkinner:
Thanks!

Issue #852, opened 2018-08-02
FluentValidation.AspNetCore should ideally use "Microsoft.AspNetCore.Mvc.Core" as a dependency vs. "Microsoft.AspNetCore.Mvc" for Web APIs


rdrenker:
### System Details

- FluentValidation.AspNetCore version: 7.6.104
- Web Framework version ASP.NET Core 2.1

### Issue Description
I'm building a Web API using ASP.NET Core;  this is a pure API and I don't require any Razor pages, or the extra overhead that Razor introduces.

As such, I'm leveraging "Microsoft.AspNetCore.Mvc.Core" instead of the "Microsoft.AspNetCore.Mvc";  doing so provides a significant reduction in the published size of the API.  (Most notably, eliminating the Razor dependencies results in exclusion of the bloated "Microsoft.CodeAnalysis.CSharp.dll" and "Microsoft.CodeAnalysis.dll" - 6+ Mb in size).

Unfortunately when I bring in FluentValidation.AspNetCore, I get this extra bloat back due to the dependency on the full "Microsoft.AspNetCore.Mvc" package.  

Would it be possible to reduce the footprint of "FluentValidation.AspNetCore" by having it rely only on "Microsoft.AspNetCore.Mvc.Core"?

Scandal-UK:
I don't get any of that bloat for a Web API project, but I reference "Microsoft.AspNetCore.App" instead of the MVC one.

Scandal-UK:
If you reference **Microsoft.AspNetCore.App** instead, you will not get the bloat.

No assets from the referenced ASP.NET Core NuGet packages are deployed with the application.

rdrenker:
PR https://github.com/JeremySkinner/FluentValidation/pull/853 created.

Was able to swap out .Mvc for .Mvc.Core;  had to bring in a couple of other small dependencies (AspNetCore.DataProtection and AspNetCore.Mvc.ViewFeatures) in order to get the builds to work and continue supporting client side validation.  Neither of the dependencies include any Razor dependencies so bloat should be eliminated.

AspNetCore unit tests all passed after the swap.

rdrenker:
@Scandal-UK **Microsoft.AspNetCore.App** still has references to various Razor assemblies which result in the same bloat, ie:

![image](https://user-images.githubusercontent.com/645718/43611530-0799a5ba-966f-11e8-8bc5-127c1051b86a.png)

Scandal-UK:
[...]

PR #853, opened 2018-08-02
Replacing Microsoft.AspNetCore.Mvc with .Mvc.Core

rdrenker:
Swapping out Microsoft.AspNetCore.Mvc to use Microsoft.AspNetCore.Mvc.Core in order to reduce bundle size;  all AspNetCore tests continue to pass after the swap.


JeremySkinner:
Closing as per #852

PR #879, opened 2018-08-24
Update to .net core 2.2

JeremySkinner:
Preview1 contains no breaking changes in the validation infrastructure. 
Edit: Preview 2 contains no breaking changes either. 
Edit: Preview 3 looks good


cwe1ss:
What's the reason for adding this additional TargetFramework? You shouldn't have to add new versions of one platform as long as you don't use any features of it. Anyone using netcoreapp2.2 can import packages that target netcoreapp2.1.

JeremySkinner:
Hi, you are correct. 

The target framework is not being changed for FluentValidation.AspNetCore.

This is solely a change to the unit test project so the tests can be run against multiple platforms.

cwe1ss:
sounds good - thx for the explanation! 👍

PR #906, opened 2018-09-27
Added argument formatting.

StefanHaug:


src/FluentValidation.Tests/MessageFormatterTests.cs
```diff
@@ -55,5 +58,89 @@ public class MessageFormatterTests {
 
 			result.ShouldEqual("bar baz");
 		}
+
+		[Fact]
+		public void Format_property_value()
+		{
```
JeremySkinner: Please can you match the project coding standards (braces at end of line)

src/FluentValidation.Tests/MessageFormatterTests.cs
```diff
@@ -55,5 +58,89 @@ public class MessageFormatterTests {
...
+				string result = formatter
+				.AppendPropertyValue(123.45)
+				.BuildMessage("{PropertyValue:#.#}");
+
+				result.ShouldEqual("123.5");
+			}
+		}
+
+		[Fact]
+		public void Understands_numeric_formats()
```
JeremySkinner: This test is currently failing, could you take a look at why that's happening?

src/FluentValidation.Tests/MessageFormatterTests.cs
```diff
@@ -55,5 +58,89 @@ public class MessageFormatterTests {
...
+		{
+			var now = DateTime.Now;
+			string result = formatter
+				.AppendArgument("foo", now)
+				.BuildMessage("{foo:g} {unknown} {unknown:format}");
+
+			result.ShouldEqual($"{(now.ToString("g"))} {{unknown}} {{unknown:format}}");
+		}
+
+		internal class TemporalCultureExchange : IDisposable
```
JeremySkinner: I don't think this `TemporalCultureExchange` class is needed - the test project already has a `CultureScope` that you should be able to use to do the same thing.

src/FluentValidation.Tests/MessageFormatterTests.cs
```diff
@@ -55,5 +58,89 @@ public class MessageFormatterTests {
...
+
+					disposedValue = true;
+				}
+			}
+
+			public void Dispose()
+			{
+				Dispose(true);
+			}
+		}
```
JeremySkinner: Please can you add tests that mix both named and numeric placeholders. Eg, a message that contains something like `"{PropertyValue:#.#} {0}"` as well as tests that include both types of placeholders formatted, eg `"{PropertyValue:#.#} {0:#.#}"`

src/FluentValidation/Internal/MessageFormatter.cs
```diff
@@ -93,21 +92,26 @@ public MessageFormatter AppendPropertyValue(object value)
 		/// </summary>
 		public Dictionary<string, object> PlaceholderValues => _placeholderValues;
 
-		protected virtual string ReplacePlaceholderWithValue(string template, string key, object value) {
-			string placeholder =  GetPlaceholder(key);
-			return template.Replace(placeholder, value?.ToString());
-		}
+		// TODO: Add alignment component
+		private Regex _keyRegex = new Regex("{([^{}:]+)(?::([^{}]+))?}"); // Matches any placeholder
```
JeremySkinner: For performance, might be better for this to be static and to use `RegexOptions.Compiled`

src/FluentValidation/Internal/MessageFormatter.cs
```diff
@@ -93,21 +92,26 @@ public MessageFormatter AppendPropertyValue(object value)
 		/// </summary>
 		public Dictionary<string, object> PlaceholderValues => _placeholderValues;
 
-		protected virtual string ReplacePlaceholderWithValue(string template, string key, object value) {
-			string placeholder =  GetPlaceholder(key);
-			return template.Replace(placeholder, value?.ToString());
-		}
+		// TODO: Add alignment component
```
JeremySkinner: What does "Add alignment component" mean?

src/FluentValidation/Internal/MessageFormatter.cs
```diff
@@ -93,21 +92,26 @@ public MessageFormatter AppendPropertyValue(object value)
...
-			switch (key) {
-				case PropertyName:
-					return "{" + PropertyName + "}";
-				case PropertyValue:
-					return "{" + PropertyValue + "}";
-				default:
-					return "{" + key + "}";
-			}
+		protected virtual string ReplacePlaceholdersWithValues(string template, IDictionary<string, object> values)
+		{
```
JeremySkinner: I have 2 concerns with this:

1) This is a breaking change for anyone that inherits from `MessageFormatter` (so this would have to be a major release and the next one isn't scheduled until next year)

2) Use of a regex for token replacement is likely to have a performance impact on high-traffic systems. 

The previous `GetPlaceholder` was high performance to account for this (this is one of the codebase's hot-spots). Would it be possible to change the approach here to only use the regex in the situation where formatted placeholders are in use? If the message doesn't have any, then fall back to the existing high-performance behaviour (which also preserves backwards compatibility for inheritors).

One approach could be:
- `BuildMessage` should detect if the `messageTemplate` contains formatted placeholders
- If not, it uses the original logic (loop over placeHolder values and call `ReplacePlaceholderWithValue`)
- If it does, then it uses the new logic. 

Thoughts?

src/FluentValidation/Internal/MessageFormatter.cs
```diff
@@ -93,21 +92,26 @@ public MessageFormatter AppendPropertyValue(object value)
...
+
+				if (!values.ContainsKey(key))
+					return m.Value; // No placeholder / value
+
+				var format = m.Groups[2].Success // Format specified?
+					? $"{{0:{m.Groups[2].Value}}}"
+					: null;
+
+				return format == null
+					? values[key]?.ToString()
```
JeremySkinner: How does this handle the situation where a placeholder exists in the template, but no value has been specified? Guessing this will throw a key not found exception. The old behaviour would leave the placeholder in the string.

Could we have a test to make sure that works correctly?

src/FluentValidation.Tests/MessageFormatterTests.cs
StefanHaug: How do I get the details from your automated build and tests? It does not fail locally.

src/FluentValidation/Internal/MessageFormatter.cs
StefanHaug: padding. https://docs.microsoft.com/en-us/dotnet/standard/base-types/composite-formatting

src/FluentValidation/Internal/MessageFormatter.cs
StefanHaug: Cannot use Compiled as fluentvalidation dependency netstandard1.1 don't have it.

JeremySkinner: Correct, but you can use it with the other target frameworks using conditional compilation, eg:

```
#if NETSTANDARD1_1 || NETSTANDARD1_6
private Regex _keyRegex = new Regex("{([^{}:]+)(?::([^{}]+))?}");
#else
private Regex _keyRegex = new Regex("{([^{}:]+)(?::([^{}]+))?}", RegexOptions | Compiled);
#endif

Let's add this please.

File: src/FluentValidation/Internal/MessageFormatter.cs (hunk 6 above)
Comment (JeremySkinner) [reply src/FluentValidation/Internal/MessageFormatter.cs]: Let's remove the todo comment. Can revisit later if needbe.

File: src/FluentValidation.Tests/MessageFormatterTests.cs (hunk 2 above)
Comment (JeremySkinner) [reply src/FluentValidation.Tests/MessageFormatterTests.cs]: Here is the test output for the tests that are failing:

Space before the percent sign. This failed under net452:
```
xUnit.net 00:00:10.0480979]     FluentValidation.Tests.MessageFormatterTests.Understands_numeric_formats [FAIL]
Failed   FluentValidation.Tests.MessageFormatterTests.Understands_numeric_formats
Error Message:
 Assert.Equal() Failure
Expected: ($123.457) ($123.4568) 1234.57 91.20% 123 1.052033e+003 1052.03 0123
Actual:     ($123.457) ($123.4568) 1234.57 91.20 % 123 1.052033e+003 1052.03 0123
```

Negative number/brackets/trailing zeros. Failed under both netcoreapp2 and netcoreapp2.1
```
[xUnit.net 00:00:08.2479321]     FluentValidation.Tests.MessageFormatterTests.Understands_numeric_formats [FAIL]
Failed   FluentValidation.Tests.MessageFormatterTests.Understands_numeric_formats
Error Message:
 Assert.Equal() Failure
Expected: ($123.457) ($123.4568) 1234.57 91.20% 123 1.052033e+003 1052.03 0123
Actual:   -$123.457 -$123.4568 1234.567 91.200% 123 1.052033e+003 1052.03 0123
```

If you're not seeing the failures locally I expect it's related to the locale. Your `TemporalCultureExchange` only set `CurrentCulture`, but doesn't change `CurrentUICulture`, which I expect is the issue. Switching to the existing `CultureScope` will probably solve this.

File: src/FluentValidation/Internal/MessageFormatter.cs (hunk 8 above)
Comment (StefanHaug) [reply src/FluentValidation/Internal/MessageFormatter.cs]: Doesn't this test do that?

		public void Should_ignore_unknown_parameters() {
			var now = DateTime.Now;
			string result = formatter
				.AppendArgument("foo", now)
				.BuildMessage("{foo:g} {unknown} {unknown:format}");

			result.ShouldEqual($"{(now.ToString("g"))} {{unknown}} {{unknown:format}}");
		}

Comment (JeremySkinner) [reply src/FluentValidation/Internal/MessageFormatter.cs]: You’re right, I missed that one.

File: src/FluentValidation/Internal/MessageFormatter.cs (hunk 5 above)
Comment (StefanHaug) [reply src/FluentValidation/Internal/MessageFormatter.cs]: I'm having trouble making the regex stuff compatible with all the frameworks. net45 and netstandard2.0 doesn't support it.

**Comment by StefanHaug** [comment]
Any ideas?

**Comment by JeremySkinner** [comment]
@StefanHaug sorry, what's your last comment referring to?

**Comment by StefanHaug** [comment]
Sorry, i try to follow your style, but i guess i overlooked another one. 

Not sure how we can add a test for the end result (black box) as we cannot really identify which implementation executed? 

I think we ought to replace the static strings with a formatted one, which uses the current framework and culture. After all, we’re not really testing the output og a string.format but rather if message builder behaves equally.

**Comment by JeremySkinner** [comment]
Latest test failure (under net452):

```
[xUnit.net 00:00:06.2461816]     FluentValidation.Tests.MessageFormatterTests.Adds_formatted_argument_and_formatted_custom_arguments [FAIL]
Failed   FluentValidation.Tests.MessageFormatterTests.Adds_formatted_argument_and_formatted_custom_arguments
Error Message:
 Assert.Equal() Failure
Expected: 123.4 67.9%
Actual:   123.4 67.9 %
```

JeremySkinner:
List of remaining issues:

- [ ] Test failure
- [ ] Code formatting (braces are still not formatted properly in several places. They should be end of line)
- [ ] `RegexOptions.Compiled` should be used on the frameworks that support it.

StefanHaug:
If there's any place I've missed the formatting I'll need you to tell me where, because I cannot see any...

JeremySkinner:
Fomatting looks good to me - thanks for making the changes. Only things I think that's outstanding is the `_templateRegex` still has `RegexOptions.Compiled` commented out (line 72 of MessageFormatter), and the commented code on line 84 which can be removed.

JeremySkinner:
Thanks, I will aim to do some further testing on this next week and if all is good I'll get this into the 8.1 release.

PR #932, opened 2018-10-28
Add EnumValidator to PersianLanguage

sohilww:
PersianLanguage.cs Doesn't have a translate for EnumValidator, I Added translate


JeremySkinner:
Thanks!

PR #938, opened 2018-11-01
Timeout for custom validator.

andrechi1:
Timeout for custom validator. 

I've put 100 milliseconds fixed. It might be better to put something configurable.

### DIFF
--- a/src/FluentValidation.Tests/RuleBuilderTests.cs
+++ b/src/FluentValidation.Tests/RuleBuilderTests.cs
@@ -155,7 +155,7 @@ namespace FluentValidation.Tests {
 			var validator = new Mock<IPropertyValidator>();
 			builder.SetValidator(validator.Object);
 
-			builder.Rule.ValidateAsync(new ValidationContext<Person>(person, new PropertyChain(), new DefaultValidatorSelector()), new CancellationToken()).Result.ToList();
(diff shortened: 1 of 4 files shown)


JeremySkinner:
Could you provide some details on the use case for this? The AsyncPredicateValidator is designed to be executed asynchronously - you should never be running it synchronously at all.

andrechi1:
My problem is that I want to do a validation on the server. And sometimes it takes too long and the phone is stopped.

CancellationToken is used next to CancellationTokenSource, in order to stop the asynchronous function. In this case it stops if it takes 100 milliseconds to validate.

JeremySkinner:
I’m not following. The change you made to AsyncPredicateValidator is to a *synchronous* method which you shouldn’t be calling.

JeremySkinner:
Just to clarify that further - you should be passing your CancellationToken (with timeout preconfigured) to the root caldiator’s ValidateAsync method. It will be propagated to all the child validators.

andrechi1:
I've changed all the CancellationToken by New CancellationTokenSource (100). Token. I don't know if somewhere is not to be done.

In My code is something similar to

RuleFor(x => x.Password).CustomAsync(async (password, ctx, cancel) =>
{
      var errors = ValidatePassword(password, cancel);

      foreach (var error in erros)
      {
            ctx.AddFailure(error.Description);
      }
}

JeremySkinner:
As I said previously, you need to specify your CancellationToken as *part of the call to ValidateAsync on your root level validator*.

andrechi1:
I've seen how do it. Thank you.

Issue #909, opened 2018-09-28
ShouldNotHaveValidationErrorFor should output validation errors


jzabroski:
This is the stack trace I get in Resharper Unit Test Session:
```
FluentValidation.TestHelper.ValidationTestException : Expected no validation errors for property Name
   at FluentValidation.TestHelper.ValidationResultTester`2.ShouldNotHaveValidationError(IEnumerable`1 properties) in C:\Projects\FluentValidation\src\FluentValidation\TestHelper\ValidationResultTester.cs:line 73
```
It should say:
```
Expected no validation errors for property Name
----
Validation Errors:
  [0]: error message 1
  [n]: error message n
```
This helps figure out which rules were triggered.

jzabroski:
https://github.com/JeremySkinner/FluentValidation/blob/253faf602dfd6103dcb0773c40392c60969415c3/src/FluentValidation/TestHelper/ValidationResultTester.cs#L61-L62

should be:
```csharp
    if (!failures.Any())
    {
        var errorMessageBanner = $"Expected a validation error for property {propertyName}\n";
        string errorMessageDetails;
        var i = 0;
        foreach (var failure in failures)
        {
            errorMessageDetails += $"\t[{i}]: {failure.ErrorMessage}\n";
        }
 	throw new ValidationTestException(errorMessageDetails);
    }
```

JeremySkinner:
Thanks for the suggestion - It’s a good idea. I’ll put it on the backlog, although it’s probably going to be a while until I can get to this.

jzabroski:
Can I submit the above as a patch, or is there more work to do?

JeremySkinner:
If you’d like to submit it as a pull request then I’d be happy to review it. I expect It’ll also require several tests to be changed too. Thanks.

JeremySkinner:
Added by PR in 472e55f.

jzabroski:
Thanks, Jeremy! Hope all is well with you.

PR #942, opened 2018-11-13
improved test exception method for ValidationResultTester.ShouldNotHaveValidationError

bdbehrho:
Pull request for "ShouldNotHaveValidationErrorFor should output validation errors #909"

### DIFF
--- a/src/FluentValidation.Tests/ValidatorTesterTester.cs
+++ b/src/FluentValidation.Tests/ValidatorTesterTester.cs
@@ -62,6 +62,12 @@ namespace FluentValidation.Tests {
 			Assert.Contains(nameof(Person.NickNames), validationTestException.Message);
 		}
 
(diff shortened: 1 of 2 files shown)


JeremySkinner:
Thanks for this, please could you add a test for the new behaviour too?

JeremySkinner:
Also, the behaviour between `ShouldNotHaveValidationError` and `ShouldHaveValidationError` is now inconsistent, so please can you ensure that `ShouldHaveValidationError` is updated to behave in the same way. Thanks.

JeremySkinner:
Thanks for adding the test. For `ShouldHaveValidationError` I meant that it might be worth listing any properties that did have error messages instead. Don't worry about this I'll add it in.

JeremySkinner:
Merged, thanks.

PR #964, opened 2018-12-05
 Add Japanese Language to the default messages

shika358:
Hi, thank you for creating a good library.
I am translating this validation message to Japanese.
Could you check this and add it to this library?


shika358:
oh, I missed it.
thank you for checking.
I fixed it.
Could you check it again?

PR #972, opened 2018-12-12
Remove netstandard2.0 package references

slang25:
Commit 6487ee5 removes the netstandard2.0 package references. Currently it can lead to `NU1605` build errors in consuming projects when another package references a lower package version. To recreate the issue, reference FluentValidator and Json.NET version 9 in a netstandard2.0 project and try to build it.

The other commits are optional.

### DIFF
--- a/src/FluentValidation/FluentValidation.csproj
+++ b/src/FluentValidation/FluentValidation.csproj
@@ -57,7 +57,7 @@ Full release notes can be found at https://github.com/JeremySkinner/FluentValida
   <ItemGroup>
     <Compile Include="..\CommonAssemblyInfo.cs" Link="CommonAssemblyInfo.cs" />
   </ItemGroup>
-  <ItemGroup>
+  <ItemGroup Condition="'$(TargetFramework)'!='netstandard2.0'">
     <PackageReference Include="System.ComponentModel.Primitives" Version="4.3.0" />
     <PackageReference Include="System.Collections.Concurrent" Version="4.3.0" />
     <PackageReference Include="System.ValueTuple" Version="4.5.0" />


slang25:
Sure, I'll send a repro over later today

slang25:
If you have the following in your `csproj`:
```xml
<Project Sdk="Microsoft.NET.Sdk">

    <PropertyGroup>
        <TargetFramework>netstandard2.0</TargetFramework>
    </PropertyGroup>
    <ItemGroup>
        <PackageReference Include="FluentValidation" Version="8.1.1" />
        <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />

        <!-- Uncomment the following 3 lines to suppress NU1605 -->
        
        <!-- <PackageReference Include="System.Text.Encoding.Extensions" Version="4.3.0" NoWarn="NU1605" />
        <PackageReference Include="System.Runtime.InteropServices" Version="4.3.0" NoWarn="NU1605" />
        <PackageReference Include="System.IO.FileSystem.Primitives" Version="4.3.0" NoWarn="NU1605" /> -->
    </ItemGroup>
    
</Project>
```

Then if you wanted to create a self-contained app with the following command: 
`dotnet publish --self-contained -r linux-x64`

You'll get downgrade warnings. As you say, mixing `netstandard1.x` and `netstandard2.x` isn't desirable, this could be fixed by upgrading `Newtonsoft.Json` or suppressing `NU1605`.

This issue came up in another OSS project, and I suggested those workarounds, however it occurred to me the package references in FluentValidation are redundant, they used to be a good idea in `netstandard1.x` but don't provide benefits in `netstandard2.x`. This PR just makes it a bit more friendly to these edge case scenarios.

slang25:
After doing some more testing, I now realise that the same downgrade warnings will happen with this PR. It has no real benefit so am closing.

slang25:
Just realised my testing was wrong, tried again with the right package that was built and it does fix the issue.

JeremySkinner:
Merged, thanks

Issue #1000, opened 2019-01-08
Addition of ScalePrecisionValidator extension method


robert-mccausland:
I can't find a relevant `ScalePrecisionValidator` extension method in the `DefaultValidatorExtensions` class, so to use the `ScalePrecisionValidator` someone needs to create the validator directly. All the other validators have one so it seems strange that this one does not.

Something like: `RuleFor(person => person.Discount).ScalePrecision(2, 7)`

Would this be something that you would want? (I can provider a PR for it if you would like).

Thanks,
Robert.

JeremySkinner:
Thanks, yes please - if you'd like to submit a PR I'll happily review it.

PR #1002, opened 2019-01-09
Add extension method for the ScalePrecisionValidator

robert-mccausland:
Added an extension method to add a `ScalePrecisionValidator` to an `IRuleBuilder`.

eg:
```csharp
validator.RuleFor(x => x.DecimalValue).ScalePrecision(2, 7);
```
#1000

### DIFF
--- a/src/FluentValidation.Tests/DefaultValidatorExtensionTester.cs
+++ b/src/FluentValidation.Tests/DefaultValidatorExtensionTester.cs
@@ -210,6 +210,19 @@ namespace FluentValidation.Tests {
 			validator.RuleFor(x => x.NullableInt).GreaterThanOrEqualTo(x => x.OtherNullableInt);
 			AssertValidator<GreaterThanOrEqualValidator>();
 		}
+
+		[Fact]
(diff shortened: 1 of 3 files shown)


JeremySkinner:
Thanks for this. Can the method also have an optional parameter for ignoreTrailingZeros? 

It’d also be good if the existing tests could be updated to go through this method.

robert-mccausland:
I have updated the Pull Request accordingly. Let me know if there is anything else.

JeremySkinner:
Please could the existing tests be updated to use the new method? At the moment they still use SetValidator directly: https://github.com/JeremySkinner/FluentValidation/blob/master/src/FluentValidation.Tests/ScalePrecisionValidatorTests.cs I think it's just lines 34 and 56 that need changing.

Also in the new method let's inline the validator class rather than using a variable:

```
ruleBuilder.SetValidator(new ScalePrecision(...) { IgnoreTrailing... = ignoretrailing... });
```

JeremySkinner:
Merged, thanks!

PR #1014, opened 2019-01-21
WIP: Add option for SqlServerCompatible to ScalePrecisionValidator

### LINKED ISSUE #1008: Add option to ScalePrecisionValidator to check digits to left of decimal point
### System Details

- FluentValidation version: 8.1.2
- Web Framework version: ASP.NET Core 2.1

### Issue Description
When using ScalePrecisionValidator to validate input that is going to be saved into MS SQL Server, the validation can pass, but then fail to save into SQL Server due to a difference in how SQL Server looks at precision and scale and how FluentValidation is using it.

If you look at the [Decimal Type documentation for SQL Server.](https://docs.microsoft.com/en-us/sql/t-sql/data-types/decimal-and-numeric-transact-sql?view=sql-server-2017) It says:

> s (scale)
The number of decimal digits that will be stored to the right of the decimal point. This number is subtracted from p to determine the maximum number of digits to the left of the decimal point.

If you look at the code for [ScalePrecisionValidator](https://github.com/JeremySkinner/FluentValidation/blob/master/src/FluentValidation/Validators/ScalePrecisionValidator.cs). It is just checking `if (scale > Scale || precision > Precision)`. It doesn't care about how many digits are to the left of the decimal point.

It would be nice to add an option so the behavior of ScalePrecisionValidator could be set up to match what SQL Server expects. This could be something like how IgnoreTrailingZeros works so you opt-in to using this different behavior. I can put up a PR for this change, but wanted to double check if it would be a good approach to this since I saw in #151 that it was just recommended to use a custom validator for this. I think it would be useful for this to be built into ScalePrecisionValidator since SQL Server is a commonly used database and its not obvious what is going wrong when you first run into this difference between the two.

Here is an example of a validator that passes, but then fails when the input goes down to SQL Server.

```
using FluentValidation;

public class TestClass
{
    public decimal Value { get; set; }
}

public class TestValidator : AbstractValidator<TestClass>
{
    public TestValidator()
    {
        RuleFor(x => x.Value).SetValidator(new ScalePrecisionValidator(4, 9) {IgnoreTrailingZeros = true});
    }
}

var testClass = new TestClass{Value = 123456};

// This passes
var result = validator.Validate(testClass);
```

However if the SQL Server `Value` column is also set to a scale of 4 and a precision of 9, inserting the value `123456` throws `System.Data.SqlClient.SqlException (0x80131904): Error converting data type numeric to decimal.`

KevinHicksSW:
Closes #1008 

One thing that needs to be figured out is how to customize the validation error message to make it clear why it fails when using the SqlServerCompatible option.

### DIFF
--- a/src/FluentValidation.Tests/DefaultValidatorExtensionTester.cs
+++ b/src/FluentValidation.Tests/DefaultValidatorExtensionTester.cs
@@ -224,6 +224,12 @@ namespace FluentValidation.Tests {
 		}
 
 		[Fact]
+		public void ScalePrecision_should_create_ScalePrecisionValidator_with_sql_server_compatible() {
+			validator.RuleFor(x => x.Discount).ScalePrecision(2, 5, true, true);
+			AssertValidator<ScalePrecisionValidator>();
+		}
+
+		[Fact]
 		public void MustAsync_should_not_throw_InvalidCastException() {
 			var model = new Model
 			{
--- a/src/FluentValidation.Tests/ScalePrecisionValidatorTests.cs
+++ b/src/FluentValidation.Tests/ScalePrecisionValidatorTests.cs
@@ -74,5 +74,53 @@ namespace FluentValidation.Tests {
 			result = validator.Validate(new Person {Discount = 00000000000000000000155.0000000000000000000000000M});
 			result.IsValid.ShouldBeTrue();
 		}
+
+		[Fact]
+		public void Scale_and_precision_should_work_with_sql_server_compatible()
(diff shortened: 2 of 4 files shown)


nhowell:
I left a comment on the related issue, as well, but wanted to voice my support for making the "sql server compatible mode" the the only way that ScalePrecisionValidator works.

JeremySkinner:
Superceded by https://github.com/FluentValidation/FluentValidation/pull/1356

PR #1041, opened 2019-02-20
Add a validator test extension for checking failure severity

Flamtap:
Added a test helper extension method for testing/asserting on validation failure Severity, similar to those for custom state, messages, and error codes.

I have an extension method in my own code to do exactly this, tying it in with the expected/actual messaging would be swell.


JeremySkinner:
Thanks for this. Could you include some test coverage too?

Flamtap:
Unit test added in c5e8e37.

JeremySkinner:
Looks good, thanks. I'll get this in the next release

PR #1045, opened 2019-02-23
DI - Add from types and custom discovery

nsarris:
Add an extension to scan directly from types since that's what AssemblyScanner inherently offers.

Also added the ability to add from a custom AssemblyScan using the AssemblyScanResult

### DIFF
--- a/src/FluentValidation.AspNetCore/FluentValidationMvcExtensions.cs
+++ b/src/FluentValidation.AspNetCore/FluentValidationMvcExtensions.cs
@@ -39,15 +39,13 @@ namespace FluentValidation.AspNetCore {
 		///     An <see cref="T:Microsoft.Extensions.DependencyInjection.IMvcCoreBuilder" /> that can be used to further configure the
 		///     MVC services.
 		/// </returns>
-		public static IMvcCoreBuilder AddFluentValidation(this IMvcCoreBuilder mvcBuilder, Action<FluentValidationMvcConfiguration> configurationExpression=null) {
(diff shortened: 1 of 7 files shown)


nsarris:
I synced with upstream, I guess I should have created a new fork. You can go ahead and drop this PR I'll create another one from a new fork. 

On the implementation side, I thought about plugging a class/func for discovery but that would create a method that took IEnumerable<Type> and a Func<IEnumerable<Type>, AssemblyScanResult> and would simple pass the types collection to the func which was pointless. So I guess just passing an IEnumerable<AssemblyScanResult> works best (after all the assembly scanner implements the same interface). The user can perform the discovery and pass the tuples.

There are a few options and discussion points here:

- Create an interface IValidatorTuple or IValidatorScanResult etc and have AssemblyScanResult implement it so that you don't have to expose it. Otherwise the class could be renamed to be used in public.

- Move AssemblyScanResult outside the AssemblyScanner class which is a breaking change (although noone should use that class anyway, could be turned to internal if an interface is provided)

- The AssemblyScanner is actually a TypeScanner with a set of static methods that extract types from assemblies.  You could move the instance implementation to a TypeScanner class (or perhaps ValidatorScanner)

JeremySkinner:
Let me have a think about this...as this isn't a common scenario, I definitely don't want to make any breaking changes, and as it's so easy to write a loop that does a scan maybe we don't need these methods at all.

(Also just a tip, you don't have to create a new fork - you can use git reset to force your fork back to the same state as upstream)

PR #1055, opened 2019-03-18
Allow registering multiple assemblies with one call

crnd:
Currently registering several assemblies requires calling `RegisterValidatorsFromAssembly(...)` several times. If you already have a list of the assemblies it would be cleaner to register the whole list of assemblies with a single call.

This PR introduces `RegisterValidatorsFromAssemblies(IEnumerable<Assembly> assemblies)` that allows you to register several assemblies with a single call.

### DIFF
--- a/src/FluentValidation.AspNetCore/FluentValidationMvcConfiguration.cs
+++ b/src/FluentValidation.AspNetCore/FluentValidationMvcConfiguration.cs
@@ -81,6 +81,15 @@ namespace FluentValidation.AspNetCore {
 		}
 
 		/// <summary>
+		/// Registers all validators derived from AbstractValidator within the specified assemblies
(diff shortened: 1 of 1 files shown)


JeremySkinner:
Hi, thanks for the contribution. We actually already added this recently as part of moving the registration methods into a separate project. (See https://github.com/JeremySkinner/FluentValidation/pull/1040/files#diff-2fdff136c508a1403be96783da65f599R33)

The next stage of this is to swap out the registration methods on FluentValidationMvcConfiguration to use these new methods instead. I haven't got around to this yet, so I'll leave this PR open as a reminder to do that but the end result will look a bit different.

JeremySkinner:
Merged, thanks.

PR #1103, opened 2019-04-25
Modify build process to add a upper max version limit

JeremySkinner:
Run PS script as part of build to add a version max for ProjectReference dependencies


JeremySkinner:
Too hacky....will wait for an official solution

PR #1109, opened 2019-04-30
new enum validator for string properties

jhiben:
Hi Jeremy,

I have noticed that there isn't any built-in validator to match a string value against the names of an enum. It's common (in web apis for example) to serialize an enum as its name instead of its underlying value (it's a lot more meaningful for the client). 

I came accross this kind of validation in one of my projects, so I thought it would be interesting to share my code with you.

### DIFF
--- a/src/FluentValidation.Tests/StringEnumValidatorTests.cs
+++ b/src/FluentValidation.Tests/StringEnumValidatorTests.cs
@@ -67,5 +67,16 @@
 			var result = _caseInsensitiveValidator.Validate(new Person { GenderString = "invalid" });
 			result.Errors.Single().ErrorMessage.ShouldEqual("'Gender String' has a range of values which does not include 'invalid'.");
 		}
+
+		[Fact]
+		public void When_enumType_is_null_it_should_throw() {
+			Assert.Throws<ArgumentNullException>(() => new TestValidator { v => v.RuleFor(x => x.GenderString).IsEnumName(null) });
+		}
+
+		[Fact]
+		public void When_enumType_is_not_an_enum_it_should_throw() {
+			var exception = Assert.Throws<ArgumentOutOfRangeException>(() => new TestValidator { v => v.RuleFor(x => x.GenderString).IsEnumName(typeof(Person)) });
+			exception.Message.ShouldEqual("The type 'Person' is not an enum and can't be used with IsEnumName.\r\nParameter name: enumType");
+		}
 	}
 }
--- a/src/FluentValidation.Tests/StringEnumValidatorTests.cs
+++ b/src/FluentValidation.Tests/StringEnumValidatorTests.cs
@@ -1,5 +1,4 @@
-﻿namespace FluentValidation.Tests
-{
+﻿namespace FluentValidation.Tests {
 	using System;
 	using System.Collections.Generic;
 	using System.Linq;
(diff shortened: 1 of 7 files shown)


JeremySkinner:
Thanks for the suggestion. I’m trying to avoid adding any new core validators as getting the error message translated into all supported language is a very big undertaking. 

Let me have a think about this.

JeremySkinner:
I thought about this yesterday...I think the best thing to do would be to use the existing error message from the `EnumValidator` rather than having a new mesasge, so we don't have to re-translate. I'll do a code review soon and then let's get this in for the next release. On initial inspection I think the main thing will be the code formatting - it needs to be consistent with the rest of the project (main thing being braces should be at the end of line).

src/FluentValidation.Tests/StringEnumValidatorTests.cs
```diff
@@ -0,0 +1,84 @@
+﻿namespace FluentValidation.Tests
+{
```
JeremySkinner: All braces should be at the end of the previous line for consistency with the rest of the project.

src/FluentValidation.Tests/StringEnumValidatorTests.cs
```diff
@@ -0,0 +1,84 @@
...
+		public void When_the_property_is_initialized_with_null_then_the_validator_should_be_valid()
+		{
+			_validator.Validate(new Person { GenderString = null }).IsValid.ShouldBeTrue();
+		}
+
+		[Fact]
+		public void When_validation_fails_the_default_error_should_be_set()
+		{
+			var result = _validator.Validate(new Person { GenderString = "invalid" });
+			result.Errors.Single().ErrorMessage.ShouldEqual("'Gender String' has names which do not include 'invalid'.");
```
JeremySkinner: Expected message will need updating to match the `EnumValidator` message.

src/FluentValidation/DefaultValidatorExtensions.cs
```diff
@@ -1085,5 +1085,32 @@ public static class DefaultValidatorExtensions {
 			action(innerValidator.RuleForEach(x => x));
 			return ruleBuilder.SetValidator(innerValidator);
 		}
+
+		/// <summary>
+		/// Defines a enum value validator on the current rule builder that ensures that the specific value is a valid enum value.
```
JeremySkinner: This will need updating to clarify that it's operating against the enum's name rather than the value.

src/FluentValidation/DefaultValidatorExtensions.cs
```diff
@@ -1085,5 +1085,32 @@ public static class DefaultValidatorExtensions {
...
+		/// <summary>
+		/// Defines a enum value validator on the current rule builder that ensures that the specific value is a valid enum value.
+		/// </summary>
+		/// <typeparam name="T">Type of Enum being validated</typeparam>
+		/// <typeparam name="TProperty">Type of property being validated</typeparam>
+		/// <param name="ruleBuilder">The rule builder on which the validator should be defined</param>
+		/// <param name="enumType">The enum whose the string should match any name</param>
+		/// <param name="caseSensitive">If the comparison between the string and the enum names should be case sensitive</param>
+		/// <returns></returns>
+		public static IRuleBuilderOptions<T, string> IsEnumName<T>(this IRuleBuilder<T, string> ruleBuilder, Type enumType, bool caseSensitive)
```
JeremySkinner: We don't need 2 overloads for this. If `caseSensitive` is made an optional parameter with a default value then we only need a single overload.

src/FluentValidation/Resources/Languages/EnglishLanguage.cs
```diff
@@ -46,6 +46,7 @@ internal class EnglishLanguage : Language {
 			Translate<EmptyValidator>("'{PropertyName}' must be empty.");
 			Translate<NullValidator>("'{PropertyName}' must be empty.");
 			Translate<EnumValidator>("'{PropertyName}' has a range of values which does not include '{PropertyValue}'.");
+			Translate<StringEnumValidator>("'{PropertyName}' has names which do not include '{PropertyValue}'.");
```
JeremySkinner: Let's remove this message and use the `EnumValidator` message instead, so re-translation isn't necessary.

src/FluentValidation/Resources/Languages/FrenchLanguage.cs
```diff
@@ -46,6 +46,7 @@ internal class FrenchLanguage : Language {
 			Translate<EmptyValidator>("'{PropertyName}' devrait être vide.");
 			Translate<NullValidator>("'{PropertyName}' devrait être vide.");
 			Translate<EnumValidator>("'{PropertyName}' a une plage de valeurs qui n'inclut pas '{PropertyValue}'.");
+			Translate<StringEnumValidator>("'{PropertyName}' a des membres qui n'incluent pas '{PropertyValue}'.");
```
JeremySkinner: Let's remove this message and use the `EnumValidator` message instead, so re-translation isn't necessary.

src/FluentValidation/Validators/StringEnumValidator.cs
```diff
@@ -0,0 +1,35 @@
...
+	{
+		private readonly Type _enumType;
+
+		private readonly bool _caseSensitive;
+
+		public StringEnumValidator(Type enumType) : this(enumType, false)
+		{
+		}
+
+		public StringEnumValidator(Type enumType, bool caseSensitive) : base(new LanguageStringSource(nameof(StringEnumValidator)))
```
JeremySkinner: Re-use the EnumValidator message with `new LanguageStringSource(nameof(EnumValidator))`

src/FluentValidation/Validators/StringEnumValidator.cs
```diff
@@ -0,0 +1,35 @@
...
+		{
+			_enumType = enumType;
+			_caseSensitive = caseSensitive;
+		}
```
JeremySkinner: We need to check that the supplied type is an enum type, which would be best done in the constructor here. Perhaps throw an exception with the message "The type '{type.Name}' is not an enum and can't be used with IsEnumName"

src/FluentValidation/DefaultValidatorExtensions.cs
jhiben: I like to separate methods into overloads with more parameters to simplify the first signature that appears within the IntelliSense, but all right for the optional parameter.

The most common way to send data over web apis nowadays is JSON, and JSON data is often camelCase. That's why I thought case insensitive by default would be less surprising. But with the optional parameter the default value will be clear, so let's go for case sensitive.

src/FluentValidation.Tests/StringEnumValidatorTests.cs
jhiben: There are several places in the code where the braces are not aligned correctly. Do you mind if I format some of them too?

JeremySkinner: Are they in files being modified by this PR? If so, then sure - that'd be great. If not then let's leave them for now, so that this PR can stay focussed on the one issue.

src/FluentValidation/Validators/StringEnumValidator.cs
```diff
@@ -0,0 +1,41 @@
...
+	using System.Reflection;
+	using FluentValidation.Internal;
+	using Resources;
+
+	public class StringEnumValidator : PropertyValidator {
+		private readonly Type _enumType;
+
+		private readonly bool _caseSensitive;
+
+		public StringEnumValidator(Type enumType) : this(enumType, false) {
```
JeremySkinner: Let's get rid of this constructor as the default is now specified as `true` in the extension method.

JeremySkinner:
I've made the following changes:
- Added license header
- Fix broken test (was failing on linux due to differences in line endings - use Environment.NewLine instead)

jhiben:
Great! And thanks for the NewLine change.

JeremySkinner:
This is now live in 8.4.0. Needs doc update at some point (#1119)

PR #1142, opened 2019-05-30
Fixed typo in word sensitive

buinauskas:
A simple typo

sensiitive ->sensitive

Also I'm not sure why last the other line is marked. Could be due to different line endings?


JeremySkinner:
Merged, thanks. Last line was because the file originally didn't have a blank newline at the end, but now does (which is correct)

PR #1148, opened 2019-06-05
Simplify TestHelper internals and advanced syntax

JeremySkinner:
The test helper "advanced" mode is syntactically messy and unclear:

```csharp
var validator = new InlineValidator<Person>();
validator.RuleFor(x => x.Surname).NotNull().WithMessage("required");
validator.RuleFor(x => x.Address.Line1).NotEqual("foo");
```

```csharp
// Current advanced test syntax
var result = validator.TestValidate(new Person { Address = new Address() });
var assert = result.Which;

assert.Property(x => x.Surname).ShouldHaveValidationError().WithMessage("required").WithErrorCode("...");
assert.Property(x => x.Address).Property(x => x.Line1).ShouldNotHaveValidationError();
```

```csharp
// New advanced test syntax
var result = validator.TestValidate(new Person { Address = new Address()) };
result.ShouldHaveValidationErrorFor(x => x.Surname).WithMessage("required");
result.ShouldNotHaveValidationErrorFor(x => x.Address.Line1);
```

- Drops the weird `Which` property
- Simplifies property syntax by moving property expression into the `Should...` methods
- Removes the need to use `Property` multiple times for nested properties
- Remove the over-complicated assertion classes internally 

Preserve backwards compat for now with deprecations. Remove in 9.0. 

- [x] needs changelog update
- [x] needs docs update


JeremySkinner:
8.5-preview2 is now released with this change.

PR #1152, opened 2019-06-11
Adding dynamic Severity

batzen:
This PR adds the ability to use a dynamic severity the same way we can use dynamic custom state by using a provider instead of a fixed value.

Use case:
We used custom state before severity existed in FluentValidation to do the same.
Our validators then created either a warning or an error depending on various conditions like "Does the user have a license to use this feature?" if he does the validation resulted in a warning, if he didn't the validation resulted in an error.

### DIFF
--- a/src/FluentValidation.Tests/UserSeverityTester.cs
+++ b/src/FluentValidation.Tests/UserSeverityTester.cs
@@ -84,5 +84,15 @@ namespace FluentValidation.Tests {
 			result = validator.Validate(person);
 			result.Errors[0].Severity.ShouldEqual(Severity.Info);
 		}
+
+		[Fact]
+		public void Should_use_last_supplied_severity() {
+			validator.RuleFor(x => x.Surname).NotNull().WithSeverity(x => Severity.Warning).WithSeverity(Severity.Info);
+
+			var person = new Person();
+
+			var result = validator.Validate(person);
+			result.Errors[0].Severity.ShouldEqual(Severity.Info);
+		}
 	}
 }
--- a/src/FluentValidation.Tests/UserSeverityTester.cs
+++ b/src/FluentValidation.Tests/UserSeverityTester.cs
@@ -77,16 +77,12 @@ namespace FluentValidation.Tests {
 
 			var person = new Person();
 
-			{
-				var result = validator.Validate(person);
-				result.Errors[0].Severity.ShouldEqual(Severity.Warning);
-			}
+			var result = validator.Validate(person);
+			result.Errors[0].Severity.ShouldEqual(Severity.Warning);
 
-			{
-				person.Age = 100;
-				var result = validator.Validate(person);
-				result.Errors[0].Severity.ShouldEqual(Severity.Info);
-			}
+			person.Age = 100;
+			result = validator.Validate(person);
+			result.Errors[0].Severity.ShouldEqual(Severity.Info);
 		}
 	}
 }
--- a/src/FluentValidation.Tests/UserSeverityTester.cs
+++ b/src/FluentValidation.Tests/UserSeverityTester.cs
@@ -1,30 +1,29 @@
 #region License
 // Copyright (c) Jeremy Skinner (http://www.jeremyskinner.co.uk)
-// 
-// Licensed under the Apache License, Version 2.0 (the "License"); 
-// you may not use this file except in compliance with the License. 
-// You may obtain a copy of the License at 
-// 
-// http://www.apache.org/licenses/LICENSE-2.0 
-// 
-// Unless required by applicable law or agreed to in writing, software 
-// distributed under the License is distributed on an "AS IS" BASIS, 
-// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. 
-// See the License for the specific language governing permissions and 
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


src/FluentValidation.Tests/UserSeverityTester.cs
```diff
@@ -44,5 +43,50 @@ public UserSeverityTester()
...
+			result.Errors[0].Severity.ShouldEqual(Severity.Warning);
+		}
+
+		[Fact]
+		public void Can_Provide_conditional_severity() {
+			validator.RuleFor(x => x.Surname).NotNull().WithSeverity(x => x.Age > 10 ? Severity.Info : Severity.Warning);
+
+			var person = new Person();
+
+			{
```
JeremySkinner: Remove these braces please - can just redefine the `result` variable.

src/FluentValidation/DefaultValidatorOptions.cs
```diff
@@ -382,6 +382,38 @@ public static class DefaultValidatorOptions {
...
+		/// </summary>
+		/// <typeparam name="T"></typeparam>
+		/// <typeparam name="TProperty"></typeparam>
+		/// <param name="rule"></param>
+		/// <param name="severityProvider"></param>
+		/// <returns></returns>
+		public static IRuleBuilderOptions<T, TProperty> WithSeverity<T, TProperty>(this IRuleBuilderOptions<T, TProperty> rule, Func<T, TProperty, Severity> severityProvider) {
+			severityProvider.Guard("A lambda expression must be passed to WithSeverity", nameof(severityProvider));
+
+			var wrapper = new Func<PropertyValidatorContext, Severity>(ctx => {
```
JeremySkinner: You can use a c#7 local function to make this nicer & easier to read.

src/FluentValidation/ValidatorMetadata.cs
```diff
@@ -74,11 +74,16 @@ public class PropertyValidatorOptions {
...
 
+		/// <summary>
+		/// Function used to retrieve the severity for the validator
+		/// </summary>
+		public Func<PropertyValidatorContext, Severity> CustomSeverityProvider { get; set; }
+
 		/// <summary>
 		/// Severity of error.
 		/// </summary>
 		public Severity Severity { get; set; }
```
JeremySkinner: Let's get rid of this entirely (see explanation in the review summary)

src/FluentValidation/DefaultValidatorOptions.cs
batzen: I just copied the code from `WithState`.
That should be changed then too i guess.

src/FluentValidation.Tests/UserSeverityTester.cs
batzen: I prefer to not reuse result variables in tests, but will change it.

src/FluentValidation/ValidatorMetadata.cs
batzen: Will do so.

src/FluentValidation/DefaultValidatorOptions.cs
batzen: To be honest, i don't fully understand what should be changed to a local function here. Could you give me a sample for this case?

JeremySkinner: I meant as an alternative to the `wrapper` variable. Not really necessary but the syntax is slightly nicer than instantiating a `Func`

```csharp
Severity SeverityProvider(PropertyValidatorContext ctx) {
	return severityProvider((T) ctx.Instance, (TProperty) ctx.PropertyValue);
}

return rule.Configure(config => config.CurrentValidator.Options.CustomSeverityProvider = SeverityProvider);
```

src/FluentValidation/ValidatorMetadata.cs
```diff
@@ -75,10 +75,10 @@ public class PropertyValidatorOptions {
 		public Func<PropertyValidatorContext, object> CustomStateProvider { get; set; }
 
 		/// <summary>
-		/// Severity of error.
+		/// Function used to retrieve the severity for the validator
 		/// </summary>
-		public Severity Severity { get; set; }
-		
+		public Func<PropertyValidatorContext, Severity> CustomSeverityProvider { get; set; }
```
JeremySkinner: Let's rename to `SeverityProvider` as there's no longer a distinction between Severity / CustomSeverity.

JeremySkinner:
If you could also rebase onto the 9.x-dev branch too that'd be great!

batzen:
> If you could also rebase onto the 9.x-dev branch too that'd be great!

Done.

src/FluentValidation/ValidatorMetadata.cs
batzen: Done

src/FluentValidation/DefaultValidatorOptions.cs
batzen: Done

JeremySkinner:
Thanks! I'll aim to get this merged in later in the week.

JeremySkinner:
Looks like there are some compile errors when retargetting to 9.0. CI didn't pick that up for some reason as azure devops was having issues and stopped picking up changes from github. I've fixed CI now, and I've pushed changes to your branch to fix the compile errors.

PR #1194, opened 2019-07-20
Added Hungarian language

alexaka1:
Added Language.cs and its constructor to LanguageManager.

### DIFF
--- a/src/FluentValidation/Resources/LanguageManager.cs
+++ b/src/FluentValidation/Resources/LanguageManager.cs
@@ -46,6 +46,7 @@ namespace FluentValidation.Resources {
 				new GeorgianLanguage(),
 				new HebrewLanguage(),
 				new HindiLanguage(),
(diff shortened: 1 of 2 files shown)


JeremySkinner:
Thanks!

alexaka1:
Hi!

May I ask when will the new translation be released into the wild? (I don't really know how this goes, I would assume an updated NuGet is required.) I'd be eager to start using it right away. I've read in another PR that you don't plan on doing any 8.x releases anymore, so will this only come out with 9.0?

JeremySkinner:
There's an 8.5 release in development which will be out before 9.0, so I'll aim to include it as part of that, but the release is tied to asp.net core 3 hitting rtm. I'll aim to publish an updated preview release with this change to nuget later this week though, if you don't mind working with pre-release versions.

JeremySkinner:
8.5-preview4 is now on nuget with this change.

PR #1221, opened 2019-09-02
Bit more helpful info when testing Validators.

### LINKED ISSUE #1220: Validating a Nullable property using ShouldHaveValidationErrorFor doesn't throw an error.
### System Details

- FluentValidation version: 8.4.0

### Issue Description

When trying to test a Nullable property (e.g. `DateTime?) that has some validation rules ... and then using `ShouldHaveValidationErrorFor` it doesn't throw an error, when it should.

The reason for this is because the value to test is really `property.Value` and this property name doesn't match the property name of the Property that errored.

E.g.

```
using FluentValidation;
using FluentValidation.TestHelper;
using System;
using Xunit;

namespace FVTest
{
    public class Tests
    {
        [Fact]
        public void GivenAnInvalidBornOn_Validate_ShouldThrowAValidationError()
        {
            var validator = new PersonValidator();

            validator.ShouldHaveValidationErrorFor(person => person.BornOn, DateTime.MinValue);
        }
    }

    public class Person
    {
        public string  Name{ get; set; }
        public DateTime? BornOn { get; set; }
    }

    public class PersonValidator : AbstractValidator<Person>
    {
        public PersonValidator()
        {
            RuleFor(person => person.BornOn.Value)
                .GreaterThan(new DateTime(2000, 1, 1))
                .When(person => person.BornOn.HasValue);
        }
    }
}
```

Run that xunit code:

image

and when I step through ....

image

image

So it's failing to see that _really_ we're doing the `.Value` property of the `Nullable` type.

PureKrome:
In #1220 I pointed out that there's a weird issue when trying to test validation rules against a `Nullable` type. Took me ages to figure out because I didn't think it was a bug with FV and only when I finally stepped through the code, did I figure out that there is an issue in FV.

This could have been mitigated _quicker_ if there was some nicer error messages. As such, I've added some more error message stuff, here.

This doesn't fix #1220 but it helps speed up time, creating tests to circumnavigate this issue.

### DIFF
--- a/src/FluentValidation.Tests/ValidatorTesterTester.cs
+++ b/src/FluentValidation.Tests/ValidatorTesterTester.cs
@@ -270,7 +270,7 @@ namespace FluentValidation.Tests {
       });
 
       var ex = Assert.Throws<ValidationTestException>(() => result.ShouldHaveValidationErrorFor(x => x.NullableInt.Value));
-      Assert.Equal("Expected a validation error for property NullableInt.Value. Found 1 error for the properties: NullableInt.", ex.Message);
+      Assert.Equal("Expected a validation error for property NullableInt.Value\n----\nProperties with Validation Errors:\n[0]: NullableInt\n", ex.Message);
     }
 
     [Fact]
@@ -287,7 +287,7 @@ namespace FluentValidation.Tests {
       });
 
       var ex = Assert.Throws<ValidationTestException>(() => result.ShouldHaveValidationErrorFor(x => x.NullableInt.Value));
-      Assert.Equal("Expected a validation error for property NullableInt.Value. Found 3 errors for the properties: NullableInt, Age, AnotherInt.", ex.Message);
+      Assert.Equal("Expected a validation error for property NullableInt.Value\n----\nProperties with Validation Errors:\n[0]: NullableInt\n[1]: Age\n[2]: AnotherInt\n", ex.Message);
     }
 
     [Fact]
--- a/src/FluentValidation.Tests/ValidatorTesterTester.cs
+++ b/src/FluentValidation.Tests/ValidatorTesterTester.cs
@@ -261,17 +261,36 @@ namespace FluentValidation.Tests {
 		}
 
 		[Fact]
-		public void ShouldHaveValidationError_should_correctly_handle_explicitly_providing_object_to_validate() {
-			var unitOfMeasure = new UnitOfMeasure {
-				Value = 1
-			};
+		public void ShouldHaveValidationError_with_an_unmatched_rule_and_a_single_error_should_throw_an_exception() {
+      var validator = new TestValidator();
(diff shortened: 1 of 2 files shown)


PureKrome:
Note: Build failed .. but I _think_ these errors are related to me?

JeremySkinner:
Thanks, but this is causing test failures as there are assertions on the error message that this PR has changed. 

- [ ] Please ensure that tests are updated that are impacted by your changes.
- [ ] In the case where there are 0 errors, the message comes out as `Expected a validation error for property Address.Line1. Found 0 errors with the properties: .` Please can this be formatted better?  We can leave off the whole  "...with the properties" section of the message if there aren't any. 
- [ ] Please can the grammar be changed? "Found x errors for the properties" rather than "found x errors with the properties".
- [ ] In fact, it might be worth including the error messages themselves. Take a look at how `ShouldNotHaveValidationError` is implemented, as this outputs the full details

PureKrome:
I'll try and see if can can build the entire solution (i had to remove most projects because of specific requirements like .NET versions, etc). I also had to hack the global.json because of missing .net core versions so i'll see if i can tweak stuff to get the stuff working agian. shouldn't be an issue.

I'll also address the other feedback - much appreciated!

JeremySkinner:
If it helps, this is what you need installed to build the project:
- .NET Core 3.0 preview 8 SDK
- .NET Core 2.2 SDK
- .NET Framework 4.5.2

PureKrome:
Is there anything .NET Core 3.0 specific? (i don't particularly want to install that).

JeremySkinner:
Yes. The asp.net core integration builds a 3.0 target.

PureKrome:
✅ Fixed grammar.
✅ Tests updated to work with new logic.
✅ New tests added.

> In fact, it might be worth including the error messages themselves. Take a look at how ShouldNotHaveValidationError is implemented, as this outputs the full details

😟  I wasn't sure what you ment by this. The code I changes is inside `ShouldNotHaveValidationError` .. so I'm confused?

🖊  Also - naming is hard. So i'm not sure if the names of the unit tests are ok. Happy to changed, of course 👍

JeremySkinner:
Thanks for the changes. I'll try and have a proper look at this later.

> I wasn't sure what you ment by this. The code I changes is inside ShouldNotHaveValidationError .. so I'm confused?

The changes you made are in ShouldHaveValidationError. If you take a look at the method below it in the class, you'll see Should*Not*HaveValidationError outputs the full details of the other error messages.

PureKrome:
Thanks for the clarification - much appreciated again!

So I've tried to make the error message more consistent now with the message from `ShouldNotHaveValidationError`. Hope this helps a bit more.

JeremySkinner:
Merged; thanks

PureKrome:
woot cheers!

PR #1235, opened 2019-09-20
Added inheritance validation

LeonardSpencer:
Added validation option for derived class in a class.

Example:

```
var validator = new TestValidator(v => {
  v.RuleFor(x => x.Vehicle.Title).NotNull();
  v.RuleFor(x => x.Vehicle).AddInheritance(i => {
    i.Include<Car>(b => { b.RuleFor(m => m.IsElectric).Must(x => x); });
    i.Include<Bike>(b => { b.RuleFor(m => m.HasBell).Must(x => x); });
  });
});

validator.Validate(new Person { Vehicle = new Bike() { HasBell = true, Title = "My awesome bike" } });

public class Person {

  (...)

  public IVehicle Vehicle { get; set; }

}

public interface IVehicle {
  string Title { get; set; }
}

public class Car : IVehicle {
  public string Title { get; set; }
  public bool IsElectric { get; set; }
}

public class Bike : IVehicle {
  public bool HasBell { get; set; }
  public string Title { get; set; }
}
```

### DIFF
--- a/src/FluentValidation.Tests/InheritanceValidatorExtensionTests.cs
+++ b/src/FluentValidation.Tests/InheritanceValidatorExtensionTests.cs
@@ -1,0 +1,28 @@
+﻿namespace FluentValidation.Tests {
+  using System.Linq;
+  using Internal;
+  using Validators;
+  using Xunit;
+
+  public class InheritanceValidatorExtensionTests {
+
+
+    private readonly AbstractValidator<Person> _validator;
+
+    public InheritanceValidatorExtensionTests() {
+      _validator = new TestValidator();
+    }
+     
+    [Fact]
+    public void AddInheritance_should_create_InheritanceValidator() {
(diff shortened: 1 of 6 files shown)


JeremySkinner:
Thank you for the suggestion, but this isn’t something that is currently on the roadmap for FluentValidation for either the 8.x series or 9.0, so this is something that will need discussion before deciding if it’s something we want to support, and if we do what the design should look like. 

Generally when contributing to open source projects it’s better to open an issue to engage the maintainers in discussion first, before going ahead and implementing new features as there are often additional design and maintenance decisions that need to be made before development can begin. 

Happy to use this PR to springboard the discussion on this, but I’m currently finishing up the 8.5 release so it’s unlikely I’ll have time to give this proper thought until after that.

LeonardSpencer:
I have created a feature request for this PR: #1237

PR #1262, opened 2019-11-17
fix site url metadata

avilv:
sharing the site via mobile etc sends the wrong link , this fixes it


JeremySkinner:
Merged, thanks

PR #1265, opened 2019-11-21
AddToModelState no longer includes a dot if PropertyName is null or empty

FTWinston:
If you have validation errors with an empty property name (because reasons), AddToModelState will still include a dot in the error key. This doesn't look great.

So if `error.PropertyName == ""`, and `prefix == "MyPrefix"`, we get this output:  `"MyPrefix."`

But it would look nicer if we instead output: `"MyPrefix"`


FTWinston:
@JeremySkinner thanks, I've updated the existing tests to account for this possibility. Have applied this to the AspNetCore, WebApi and MVC5 versions, as I'm hoping for another 8.x release :)

Issue #1286, opened 2019-12-27
DI


managerger:
I've faced an issue, while trying to set a validator for collection of objects. I inject a validator for each instance in the following way:
`RuleForEach(m => m.Users).InjectValidator((sp, cntx) => sp.GetService<BaseUserValidator<BaseUser>>());`
But when I use `RuleForEach(m => m.Users).InjectValidator((sp, cntx) => new BaseUserValidator<BaseUser>());` then everything works just fine. My best guess was that framework could'nt register `BaseUserValidator<T>`, because it doesn't work with open generics. Do we have a way to fix that?

JeremySkinner:
Hi, this isn’t really related to FluentValidation. It just means you havent registered it with the service provider, which you should do in your application’s ConfigureServices method. Just register it as an open generic and then you’ll be able to resolve it as necessary.

managerger:
#1287

JeremySkinner:
FluentValidation intentionally only registers validators with one to one mappings as the simple use case. 

Open generics are much less clear cut as it is not obvious what the default behaviour should be for these as they can be used in many different ways (which then becomes more complex when you take into account multiple type parameters, type constraints and covariant/contravariant constraints in genetics). Because of this FluentValidation intentionally does not handle them out of the box - it is your responsibility as the developer to express your intent how these should be used by registering them in the way that is most appropriate for your application. 

FluentValidation won’t be changed to auto register these.

managerger:
Yeah, now I see your point. That make sense. Thanks)

PR #1287, opened 2019-12-27
Open generics registration

managerger:
PR for #1286 
Well, I took a glance through source code and found out that FluentValidation takes care of registration. Therefore it could be its responsibility to handle open generics, as it does with regular types through `.AddFluentValidation(fv => fv.RegisterValidatorsFromAssemblyContaining<Startup>())` extension. Said that, I made some changes to the `FluentValidation` and `FluentValidation.DependencyInjectionExtension` projects and seems like it does a trick. The only thing I need is check and verify if these changes are valid, because I didn't have a much time to check everything out and honestly it is a first time a looked at source code. Thanks)


JeremySkinner:
Thanks for the suggestion, but this is not in scope for the project, so isn’t something that will be included. Handling open generics by default is not as simple as the change you’ve made - Please see my comment on the other issue for more details. Thanks.

PR #1326, opened 2020-02-19
Create SlovenianLanguge

bostjankosi:
Please add slovenian languge


JeremySkinner:
Thanks for this - you’ll also need to add it to the LanguageManager here: https://github.com/FluentValidation/FluentValidation/blob/master/src/FluentValidation/Resources/LanguageManager.cs#L47

JeremySkinner:
Please put the changes in the same pull request, not two separate ones.

PR #1338, opened 2020-03-03
add ValidationException constructor with support for custom message and errors message

bombek92:


src/FluentValidation/ValidationException.cs
```diff
@@ -49,11 +49,23 @@ public class ValidationException : Exception {
...
-		/// <param name="errors"></param>
-		public ValidationException(IEnumerable<ValidationFailure> errors) : base(BuildErrorMessage(errors)) {
+
+    /// <summary>
+    /// Creates a new ValidationException
+    /// </summary>
+    /// <param name="message"></param>
+    /// <param name="errors"></param>
+    /// <param name="addErrorMessage"></param>
+    public ValidationException(string message, IEnumerable<ValidationFailure> errors, bool addErrorMessage)
```
JeremySkinner: Could the new parameter be renamed? I think something like `appendDefaultMessage` would be slightly clearer

src/FluentValidation/ValidationException.cs
```diff
@@ -49,11 +49,23 @@ public class ValidationException : Exception {
...
-		/// </summary>
-		/// <param name="errors"></param>
-		public ValidationException(IEnumerable<ValidationFailure> errors) : base(BuildErrorMessage(errors)) {
+
+    /// <summary>
+    /// Creates a new ValidationException
+    /// </summary>
+    /// <param name="message"></param>
+    /// <param name="errors"></param>
+    /// <param name="addErrorMessage"></param>
```
JeremySkinner: Maybe include a description of the parameter?

bombek92:
> Thanks for this, a couple of small tweaks. Could you also add a test that covers the new behaviour? Thanks!

Thanks for suggestions, I've implemented them. Unit tests were added as well.

Issue #1339, opened 2020-03-03
Better handling of indexers in TestHelper


JeremySkinner:
At the moment, indexers aren't supported in the TestHelper extensions that accept lambdas, only the string-based overloads. Update the lambda-based overloads to support them too. 

The simplest thing to do would be to call ToString on the expression tree and then manipulate the string:
- Remove the `=>` and anyting before it
- Rewrite indexer calls to `get_Item(0)` to `[0]`
- Would mean having to explicitly write special cases for any other weird stuff that people might put inside expression trees (which would get very complicated)

Other option would be to use an expression-parsing library like https://github.com/AgileObjects/ReadableExpressions or https://github.com/zspitz/ExpressionTreeToString but this would mean taking a package dependency, and I'm not really sure it's worth brining along a whole other library whenever FV is installed just for the TestHelper.

jzabroski:
Can you break out TestHelper into TestHelper.Supercharged as an external lib?  Then people can choose to whether to take the dependency.

As far as manipulating expression trees goes, I found @fsoikin's erecruit.Expr to be very useful/powerful.  However, it sounds like your use case is to pretty print the expression, rather than manipulate it?

JeremySkinner:
No, I don't see any benefit to moving this to a separate library. The test helpers are tiny. An additional dependency doesn't serve any purpose here. 

I've already implemented the necessary expression tree manipulation in the PR #1340.

PR #1340, opened 2020-03-03
Support indexers in property chain expressions.

JeremySkinner:
This is to support #1339

### DIFF
--- a/Changelog.txt
+++ b/Changelog.txt
@@ -24,6 +24,7 @@ Add additional overload of SetValidator that takes a Func that receives the curr
 Work around a bug in ASP.NET Core's integration testing components that can cause ConfigureServices to run multiple times.
 SourceLink integration.
 {CollectionIndex} placeholder can now be accessed in child validators.
+Support indexers in test helper expressions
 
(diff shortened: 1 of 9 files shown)


Luke-TE:
Hi, 
is there a reason WHY this was closed without being merged?
Thanks

JeremySkinner:
I decided against implementing it as the issue is an uncommon edge case, and I didn't like the solution and don't want to maintain it.

Luke-TE:
I see. That's unfortunate, I just ran into that edge case :/

JeremySkinner:
Use the string-based overloads in the test helper rather than the lambda overloads if you need indexers.

Luke-TE:
Thanks, that works great!
I originally tried it with the `validator.ShouldHaveValidationErrorFor(...)` style but it seems to only be supported for the `result.ShouldHaveValidationErrorFor(...)` style.

JeremySkinner:
Yes, the `result.ShouldHaveValidationErrorFor` style is the way to go. The old versions will be deprecated in the next release

PR #1347, opened 2020-03-12
Support setup max number of errors

blowin:


JeremySkinner:
Hi there, please could you explain what this is for? 

 It’s usually best to open an issue to discuss feature requests before sending a pull request to ensure it fits with the project roadmap. 

It’d be helpful to have full details of your use case and what you’re trying to achieve (and details of why the existing extensibility points  don’t meet your needs). Thanks.

blowin:
Limit the number of validations over an entity. Suppose there are several validators and they are combined using include. Among them, there may be complex validations that require access to the database. In this case, performance problems may occur due to invoking all checks. There is currently no way to specify the maximum number of validations.

JeremySkinner:
This feels like a very specific use-case that'd you'd probably be better off keeping in your own project (by having a base-class for your validators) rather than something we'd have in the core library. But thank you for the suggestion.

PR #1349, opened 2020-03-18
Use the correct word for size in Dutch

CumpsD:
`grote` is like `a big house`, an adjective
`grootte` is like the size you mean here, a noun, `the size of a number`

Reference from the official language book for the Dutch language:
https://taaladvies.net/taal/advies/vraag/353/grote_grootte/

https://translate.google.com/#view=home&op=translate&sl=nl&tl=en&text=Correct%20is%3A%20een%20model%20op%20ware%20grootte.%20Grootte%20is%20een%20zelfstandig%20naamwoord%2C%20grote%20is%20een%20bijvoeglijk%20naamwoord%20(het%20grote%20model).


JeremySkinner:
Merged, thanks!

PR #1359, opened 2020-04-02
Docs: Move Some Examples to Improve Flow

### LINKED ISSUE #1211: [Docs] [Improvement] Customer Model under Collection
image
I think the `Collection` section could be improved. Under the `Collection` section I would expect all code references to revolve around `Person`. This section currently references both `Person` and `Customer`. `Customer` is actually introduced with a code snippet under `Collections of Complex Objects` which is after the `Collection` section.

Jacksondr5:
Reorder some of the examples in the Collections doc page.  There were a couple of code snippets that used classes defined later in the page, so I moved those examples below the class definition snippets.

This should close #1211

### DIFF
--- a/docs/collections.md
+++ b/docs/collections.md
@@ -17,7 +17,8 @@ public class PersonValidator : AbstractValidator<Person> {
   }
 }
 ```
-The above rule will run a NotNull check against each item in the `AddressLines` collection. 
+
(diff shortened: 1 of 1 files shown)


JeremySkinner:
Thanks!

PR #1365, opened 2020-04-13
Add ValidationContext to When predicate

dcabines:
My company is using FluentValidation to validate timesheet data, but the current methods are limiting us. We are populating the `RootContextData` with relevant information, but we cannot access it from within the `When` and `WhenAsync` methods.

We would like to do something like this:
```
When(
  (timesheetDetailEntry, context) => {
    // use context info to decide if a set of rules should run. example:
    var formType = (FormType)context.ParentContext.RootContextData[TimesheetContextKeys.FormType];
    return formType == "target form type";
  },
  () => {
    // set of rules to run
  }
);

RuleFor(timesheet => timesheet.Property)
  .NotNull()
  .WhenAsync(async (property, context, cancel) => {
    // use context info to decide if rule should run. example:
    var timesheet = (Timesheet)context.ParentContext.RootContextData[TimesheetContextKeys.Timesheet];
    var user = await UserClient.GetUserAsync(timesheet.userId);
    return user.Type == "target user type";
  });
```

### DIFF
--- a/src/FluentValidation.Tests/RuleBuilderTests.cs
+++ b/src/FluentValidation.Tests/RuleBuilderTests.cs
@@ -101,7 +101,12 @@ namespace FluentValidation.Tests {
 
 		[Fact]
 		public void Should_throw_when_predicate_is_null() {
-			typeof(ArgumentNullException).ShouldBeThrownBy(() => builder.SetValidator(new TestPropertyValidator()).When(null));
+			typeof(ArgumentNullException).ShouldBeThrownBy(() => builder.SetValidator(new TestPropertyValidator()).When((Func<Person, bool>)null));
+		}
+
+		[Fact]
+		public void Should_throw_when_context_predicate_is_null() {
+			typeof(ArgumentNullException).ShouldBeThrownBy(() => builder.SetValidator(new TestPropertyValidator()).When((Func<Person, ValidationContext<Person>, bool>)null));
 		}
 
 		[Fact]
@@ -110,8 +115,13 @@ namespace FluentValidation.Tests {
 		}
 
 		[Fact]
+		public void Should_throw_when_inverse_context_predicate_is_null() {
+			typeof(ArgumentNullException).ShouldBeThrownBy(() => builder.SetValidator(new TestPropertyValidator()).Unless((Func<Person, ValidationContext<Person>, bool>)null));
+		}
+
+		[Fact]
 		public void Should_throw_when_inverse_predicate_is_null() {
-			typeof(ArgumentNullException).ShouldBeThrownBy(() => builder.SetValidator(new TestPropertyValidator()).Unless(null));
+			typeof(ArgumentNullException).ShouldBeThrownBy(() => builder.SetValidator(new TestPropertyValidator()).Unless((Func<Person, bool>)null));
 		}
 
 		[Fact]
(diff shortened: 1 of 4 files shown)


JeremySkinner:
Hi, thanks for the suggestion. I think this is a good idea, but will need a couple of changes. I'll try and do a more thorough review later in the week but an initial couple of things that stand out to me:

- The callback should take a strongly-typed `ValidationContext<T>` rather than an untyped `ValidationContext`.
- Lets avoid duplicating the When/Unless methods in the ConditionBuilder. The logic can go into the overload that has the ValidationContext, and the overload without can just act as a wrapper for this by doing something like this:

```csharp
public IConditionBuilder When(Func<T, bool> predicate, Action action) {
   return this.When((x, context) => predicate(x), action);
}
```

If you could get these changes in, then I'll do a full review.

Thanks again for contributing!

dcabines:
I removed the duplicated methods. When I made the overload wrap the new version I saw it was only used in one place in `AbstractValidator` so I removed it from `ConditionBuilder` and had `AbstractValidator` take care of doing `(x, ctx) => predicate(x)`.

I made them use the generic `ValidationContext<T>`, but to avoid having to change too many places I am using `context.ToGeneric<T>()` within the `When` and `WhenAsync` methods.

JeremySkinner:
Calling `context.ToGeneric` actually creates a new `ValidationContext<T>`, which I don't think we need to do here. I believe should be able to just cast the existing `ValidationContext` to a `ValidationContext<T>` without any issues, which will avoid a new allocation.

It also looks like the indentation has been changed in the `When` and `WhenAsync` methods of ConditionBuilder. Could you reset this back to using tabs, for consistency with the rest of the project?

Thanks!

dcabines:
- Changed spaces to tabs
- Casting instead of calling `ToGeneric`
- Added a few overloads to `DefaultValidatorOptions`
- The old `AbstractValidator` methods call the new ones.

dcabines:
Oh, but two tests now have an ambiguous reference to `When`. I'm not sure what I should do for those. The ambiguous reference makes passing `null` impossible, but you may not want to rely on that.

JeremySkinner:
You can cast the null to a specific type to make the invocation unambiguous

dcabines:
Perfect, thanks for the tip. I added two more tests to cover the new methods.

JeremySkinner:
Thanks, I'll aim to do a final review over the weekend and then get it merged

JeremySkinner:
Oh there's a test failure too - if you could get that sorted first that'd be great. Thanks!

dcabines:
I added a missing applyConditionTo and that fixed the test, but the Windows check is complaining about some package reference.
```
Unable to find package Microsoft.NETFramework.ReferenceAssemblies
Unable to find package Microsoft.SourceLink.GitHub
```

JeremySkinner:
Thanks.  

Looks like something changed inside the windows build image. Probably
Microsoft changed something in azure.  Don’t worry about that, I’ll look at it and see if I can figure out what’s going on.

JeremySkinner:
Merged, thanks

JeremySkinner:
One minor thing I changed was I'd forgotten there's a method on ValidationContext<T> for converting a non-generic conext. It does a cast if it can, otherwise it creates a new one: https://github.com/FluentValidation/FluentValidation/blob/master/src/FluentValidation/ValidationContext.cs#L81

I switched out the casts to use this, even though they should be fine it makes them a bit more resilient.

PR #1385, opened 2020-05-09
Cleanup for out ctors rules

volodin-danila:
Very useful when you need to fetch the rules from the constructor and have instanceToValidate.
Clean Constructor is a good practice for Dependency Management.
Conditional logic can now be configured with standard statements: if-else-switch-etc.

### DIFF
--- a/src/FluentValidation.Tests/CleanupRulesTests.cs
+++ b/src/FluentValidation.Tests/CleanupRulesTests.cs
@@ -1,0 +1,51 @@
+﻿using FluentValidation.Results;
+using System;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace FluentValidation.Tests {
+  public class CleanupRulesTests {
+
+    [Fact]
+    public void Cleanup_for_settings_rules_without_ctor_with_instanceToValidate() {
+      var validator = new PersonValidatorWithotCtor();
+
+      Assert.True(validator.Validate(new Person()).IsValid);
+      Assert.False(validator.Validate(new Person { Surname = "A" }).IsValid);
+      Assert.True(validator.Validate(new Person { Surname = "A", Email = "@" }).IsValid);
+    }
+  }
+
+  public class PersonValidatorWithotCtor : BaseValidator<Person> {
+    public override async Task Configure(Person instanceToValidate) {
+
+      if (instanceToValidate.Surname?.StartsWith("A") == true) {
+        RuleFor(x => x.Email)
+            .NotEmpty();
+      }
+
+      await Task.CompletedTask;
+    }
+  }
+
+  // Very useful when you need to fetch the rules from the constructor and have instanceToValidate
+  // Clean Constructor is a good practice for Dependency Management
+  // Conditional logic can now be configured with standard statements: if-else-switch-etc
+  public abstract class BaseValidator<T> : AbstractValidator<T> {
+
+    private readonly Action<T> _action;
+
+    protected override bool PreValidate(ValidationContext<T> context, ValidationResult result) {
+
+      CleanupRules();
+
+      _action(context.InstanceToValidate);
+
+      return base.PreValidate(context, result);
+    }
+
+    public BaseValidator() => _action = new Action<T>(async x => await Configure(x));
+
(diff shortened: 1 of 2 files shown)


JeremySkinner:
Hello,

Thank you for your submission. Generally it's better to open a new issue to discuss feature requests with the project maintainers before opening the pull request to make sure that your feature fits in with the goals and direction of the project.

Validator instances are intended to be immutable once constructed - modifying validators once they've been created is not supported or recommended. Once a validator is constructed, its rules shouldn't be modified. 

If you need rules to be applied conditionally, then you should make use of features such as Conditions (the `When\Unless` methods) and `DependentRules`), but you shouldn't be using standard conditional logic (if/else etc) to control the construction of rules.

That being said, if you *really* want to go down this route then there's nothing stopping you - you can define your own validator base class and add the ClearRules method to this, but again I would highly recommend that you don't do this as it really isn't the intended use of the library. 

If you'd like to discuss this further then feel free to open a new issue to discuss this, but as this change goes against the intended immutability of validators, this isn't something we'll be accepting into the library I'm afraid.

Thanks for wanting to get involved though!

volodin-danila:
Hi, Jeremy! I can’t make my base validator, since Rules is internal. Otherwise, it will be necessary to write reflection...

`
public abstract class BaseValidator<T> : AbstractValidator<T>
{
    private readonly object _lock = new object();
    private readonly Action<T> _action;

    protected override bool PreValidate(ValidationContext<T> context, ValidationResult result)
    {
        lock (_lock)
        {
            var rules = (TrackingCollection<IValidationRule>)GetType()
                .GetProperty("Rules", BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.Public)
                .GetValue(this);

            while (rules.Count != 0)
                rules.Remove(rules.Last());
        }

        _action(context.InstanceToValidate);

        return base.PreValidate(context, result);
    }

    public BaseValidator() => _action = new Action<T>(async x => await Configure(x));

    public abstract Task Configure(T instanceToValidate);
}
`

JeremySkinner:
Hello again,

Did you fully read my previous reply about why this isn't a good idea to do, and why it's not thread-safe?

Please provide a *full and clear explanation of what you're trying to achieve*.

volodin-danila:
Yes of course. I read your message carefully. But this Idea is thread safe because it uses rules lock until cleanup. I try to achieve the following: removing rules from the constructor. Placing logic in a constructor is not a good idea and bad practice. Secondly, When \ Unless methods are very inconvenient. The standard If-else-switch branching mechanisms are much more efficient and convenient designs in complex scenarios.

JeremySkinner:
> But this Idea is thread safe because it uses rules lock until cleanup

No, you are incorrect This is not thread safe. You are not locking other readers/writers of the `Rules` collection. Calling `lock(Rules)` does not prevent other readers/writers from accessing the Rules collection - that isn't how the lock statement in C# works. It only prevents multiple callers from calling `ClearRules`, but it doesn't prevent other readers from enumerating the collection, or from other writers from adding new things to it. 

> I try to achieve the following: removing rules from the constructor.

Why? The *whole point* of FluentValidation is to define rules in the constructor. If you don't want to do this then you shouldn't be using FluentValidation!

> Placing logic in a constructor is not a good idea and bad practice

Nonsense.

>  Secondly, When \ Unless methods are very inconvenient. The standard If-else-switch branching mechanisms are much more efficient and convenient designs in complex scenarios.

Then I'll say this again: Why are you using FluentValidation at all? If you don't like the library's features, then go and use something else.

volodin-danila:
Each reader will wait in line until a personal PreValidate is completed for him. In PreValidate, rules are blocked to correctly re-initialize for a specific caller.

About the logic in the constructor https://dzone.com/articles/fixing-the-constructor-anti-pattern

I use FluentValidation because it is a validator in the world )

volodin-danila:
Ok, thank you for your comments. It will be necessary to think again about this. Perhaps taking logic out of the FluentValidation constructor is really not a good idea. Although admitting it would be very convenient to configure fluent rules through native if else etc. Perhaps you will consider such a possibility in future versions. Thanks, Jeremy!

PR #1400, opened 2020-05-25
add rollforward to SDK version

janschreier:
the rollforward makes life easier for opening the project in VS and with limiting it to latestpatch it should not create any unexpected side effects


JeremySkinner:
Thanks!

PR #1403, opened 2020-05-25
fixed broken URLs

janschreier:
fixed all links that had a wrong / 

(sorry the global.json bit is still included - I guess I should do my git course now...)


JeremySkinner:
Are you able to rebase this, as it has conflicts?

janschreier:
ok, will do, but give some time (I thought that rebase would be part of the answer, but having never really worked with Git, I have to figure out how to do this)

JeremySkinner:
Ok no problem, leave it to me I'll sort it.

JeremySkinner:
I've rebased your fork so it has a linear history without a merge commit, so this will apply cleanly now.

PR #1406, opened 2020-05-25
For Consideration: FluentAssertions

SeanKilleen:
A quick demo in case you wanted to move toward using FluentAssertions in your tests rather than custom extensions. 

It's a dependency, so I totally understand if you don't want to take that on, but it's a great library that enables a fluent syntax and has really nice outputs and a ton of built in extensions. Plus I figure it's only a dependency in your test project, so you're not shipping it.

If you're interested in this, I can continue on another time; if you're not interested, feel free to close and I won't be offended. :)

### DIFF
--- a/src/FluentValidation.Tests.AspNetCore/ClientsideMessageTester.cs
+++ b/src/FluentValidation.Tests.AspNetCore/ClientsideMessageTester.cs
@@ -21,6 +21,7 @@
 namespace FluentValidation.Tests.AspNetCore {
 	using System.Net.Http;
 	using System.Threading.Tasks;
+	using FluentAssertions;
 	using Xunit;
 
 	public class ClientsideMessageTester : IClassFixture<WebAppFixture> {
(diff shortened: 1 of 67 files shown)


JeremySkinner:
Hmm I’m not sure that I really see a benefit to this? What’s the advantage?

SeanKilleen:
The main advantage I see is additional flexibility in your assertions for future tests (great assertions against collections, etc. for example) and off-loading something custom to a library built for that purpose. Plus I like to see .NET projects utilizing other libs in the ecosystem so that's just a "warm and fuzzies" advantage for me. :) 

Likely not a priority at all; just occurred to me so thought I'd suggest.

JeremySkinner:
I've had a think about this and I'm not keen on making this changes. If it ever gets to a point where it feels like this is needed then I'd be willing to go for it, but at the moment I don't think this brings any value and I'm not keen on making changes just for the sake of it. Thanks for the suggestion though!

Issue #820, opened 2018-06-30
Documentation updates


JeremySkinner:
- [x] Missing validators on the validator list page.
- [ ] ~~Custom message builders~~ Edit: This is really hard to document, so I think I'll leave it as a hidden advanced feature
- [x] Advanced test extensions
- [x] Async
- [x] Make cascade mode clearer 
- [x] Note about root page model with aspnet web pages (as per #887)
- [x] `Otherwise`
- [x] Why the email validator is intentionally not 100% comprehensive, and why this is OK.
- [x] Make it more obvious that InjectValidator is only available when using automatic validation. 
- [x] Severity
- [x] State
- [x] ErrorCode

MajeQafouri:
Thank you

benmccallum:
Hi @JeremySkinner, as mentioned over on #999 you had it on the list to doco async PropertyValidator usage, but I don't see it here unless it was covered by the "Async" item which is checked but I've been unable to find anything on the docs site under [PropertyValidator](https://fluentvalidation.net/custom-validators#reusable-property-validators).

I wanted to get some confirmation that using the technique on #999 (putting GetAwaiter().GetResult() in the IsValid method) wouldn't cause blocking. My knowledge in this area would suggest it could lead to deadlocks, but I'd love to be proven wrong as otherwise I'm not sure how to workaround this right now.

JeremySkinner:
@benmccallum please open a separate issue explaining clearly what you’re trying to achieve and the problem you’re running into. This issue is a todo list for doc updates I need to write.

SeanKilleen:
@JeremySkinner since we've checked off Severity, State ,and error codes, do you want to provide a little insight into what you mean by custom message builders? If so, I can run with it.

SeanKilleen:
Also, should this have a 9.0 milestone?

JeremySkinner:
> Also, should this have a 9.0 milestone?

No these are just ongoing doc updates, not necessarily related to 9. the ones that are are already in the 9.0 milestone #1405 and #981 

> do you want to provide a little insight into what you mean by custom message builders

Yes its an internal API feature that allows users to completely customize the process of generating the error message, and can be set using the `MessageBuilder` property on the rule. Here's an example of how i've suggested using it before: 

https://github.com/FluentValidation/FluentValidation/issues/1353#issuecomment-604951897

PR #1408, opened 2020-05-25
Adding docs: Severity, ErrorCode, CustomState

SeanKilleen:
Supports #820.

I tentatively added these docs to a topic called "Validation Metadata", because I found it to be data that is tangential to validation. I'm not sure metadata is the correct term.

Happy to have it moved anywhere, and it's a first pass so blunt feedback and direct edits are welcome. :)


JeremySkinner:
Thanks! I’m not going to have time to properly review this today but will try and do so later in the week.

JeremySkinner:
I've made a few changes:
- Renamed to "Additional options". Metadata is already a thing that exists (ie obtaining metadata from validator.CreateDescriptor) and don't want this to be ambiguous
- Reworded a few bits and some grammar fixes
- Added output examples

What's left to do:
- [ ] The section on error codes needs to talk about how error codes affect error message lookups in the LanguageManager
- [ ] Needs linking to the TOC or it won't appear in the menu. 

I'm happy to do these

SeanKilleen:
@JeremySkinner I'm happy to do the remaining items but if you're on a roll feel free to knock it out. 👍

SeanKilleen:
@JeremySkinner I think this is now ready for review when you're able. No rush.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
JeremySkinner, line 1: Don't think we need this file?

docs/additional-options.md
```diff
@@ -0,0 +1,105 @@
...
+The output would be:
+
+```
+Property: Surname Error Code: ERR1234
+Property: Forename Error Code: NotNullValidator
+```
+
+Note that if an error code is not explicitly specified, it will default to the name of the underlying property validator (eg for a call to `NotNull` it would be `"NotNullValidator"`).
+
+Note that this affects the localized lookups of error messages as well, as the error code is used for the localization key. If you override the error code in a localized environment, you'll need to ensure that you provide a localized value using that error code as well. When there is a custom error code with no key, FluentValidation falls back to the standard key. For example, if you provide an error code of `ERR123` on a `NotNull` validation, you'll want to have a localization value by `ERR123` in your resource file. If you do not supply a translation, the resource for `NotNullValidator` will be used in its place.
```
JeremySkinner, line 76: > you'll need to ensure that you provide a localized value using that error code as well

Maybe I'm misreading this, but that's not the case. If there's no localized lookup using the error code it still uses the default.

JeremySkinner: Ie you can do `.Must(x => ...).WithErrorCode("NotNullValidator")` to inherit the error message for the notnullvalidator

SeanKilleen: I see -- I misunderstood what we were trying to accomplish with that. I'll fix it up. 👍

SeanKilleen:
@JeremySkinner I've tried to make this more succinct and use bullet points instead. Let me know if that works.

.vscode/settings.json
JeremySkinner: I'll go ahead and delete this file.

JeremySkinner:
Thanks, merged. I made a few tweaks to the wording and also decided to split it up into 3 separate pages as I think these are big enough topics that they deserve their own menu items

SeanKilleen:
Happy to help! Thanks for all the assistance.

PR #1456, opened 2020-07-21
Add IActionContextValidatorInterceptor and avoid throwing if the ActionContext is not a Controller Context

david-driscoll:
Related issue: https://github.com/FluentValidation/FluentValidation/issues/1457

Update:
Based on the discussion in #1457 I have updated this PR with the following:

A new interface `IActionContextValidatorInterceptor` which is already marked as deprecated to ensure IValidatorInterceptor gets updated next major version to use `ActionContext` instead of `ControllerContext`
Ensure that if a `IValidatorInterceptor` is registered globally it does not cause an runtime exception for Razor Pages

I did my best to add appropriate tests that validate the correct behaviors.

### DIFF
--- a/src/FluentValidation.AspNetCore/CustomizeValidatorAttribute.cs
+++ b/src/FluentValidation.AspNetCore/CustomizeValidatorAttribute.cs
@@ -1,4 +1,4 @@
-﻿#region License
+#region License
 // Copyright (c) .NET Foundation and contributors.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
@@ -95,6 +95,7 @@ namespace FluentValidation.AspNetCore {
 			if (Interceptor == null) return null;
 
 			if (!typeof(IValidatorInterceptor) .GetTypeInfo().IsAssignableFrom(Interceptor)) {
+				if (typeof(IActionContextValidatorInterceptor).GetTypeInfo().IsAssignableFrom(Interceptor)) return null;
 				throw new InvalidOperationException("Type {0} is not an IValidatorInterceptor. The Interceptor property of CustomizeValidatorAttribute must implement IValidatorInterceptor.");
 			}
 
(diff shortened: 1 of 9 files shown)


src/FluentValidation.AspNetCore/CustomizeValidatorAttribute.cs
```diff
@@ -1,4 +1,4 @@
-﻿#region License
+#region License
```
david-driscoll, line 1: I think this was the BOM... I was using Github Codespaces to do the work and couldn't find a way to restore it.  Let me know if this is bothersome and I'll pull it down on my machine and fix it.

src/FluentValidation.AspNetCore/FluentValidationModelValidatorProvider.cs
```diff
@@ -81,23 +81,30 @@ public class FluentValidationModelValidator : IModelValidator {
...
 				IValidationContext context = new ValidationContext<object>(mvContext.Model, new PropertyChain(), selector);
 				context.RootContextData["InvokedByMvc"] = true;
 				context.SetServiceProvider(mvContext.ActionContext.HttpContext.RequestServices);
 
-				if (interceptor != null) {
+				if (interceptor != null && mvContext.ActionContext is ControllerContext) {
 					// Allow the user to provide a customized context
 					// However, if they return null then just use the original context.
 					context = interceptor.BeforeMvcValidation((ControllerContext)mvContext.ActionContext, context) ?? context;
+				} else if (actionContextInterceptor != null) {
```
david-driscoll, line 96: Let me know if you think the logic makes sense.

In the event there are two interceptors registered, the existing one takes precedence, unless the `ActionContext` is not a `ControllerContext` then the other interceptor will pick up.

src/FluentValidation.AspNetCore/CustomizeValidatorAttribute.cs
JeremySkinner: That's fine, some of the older files in the repo (like this) have some minor formatting issues from before I added an `.editorconfig` file. GH is just correcting this :)

src/FluentValidation.AspNetCore/FluentValidationModelValidatorProvider.cs
JeremySkinner: Looks good

JeremySkinner:
Looks good, thanks! I've made some very minor formatting tweaks and will merge once CI passes

PR #1505, opened 2020-08-28
support netstandard2.0 and remove razor dependency

dhwed:
We are in the process of migrating to .NET 5, but right now we are still using AspNetCore 2.1 on .NET Framework, and we would like to be able to continue to use the latest versions of FluentValidation.AspNetCore, so we still need the `netstandard2.0` target.

In addition, the razor dependencies in `Microsoft.AspNetCore.Mvc` bring in several additional megabytes of unused binaries to each of our applications.  This library does not require any of the features for cshtml file compilation, so we can change the dependency to `Microsoft.AspNetCore.ViewFeatures`.

Note that there will be a conflict with #1474 (you'll have to make a minor change to the conditional compilation in one place).

### DIFF
--- a/src/FluentValidation.AspNetCore/FluentValidation.AspNetCore.csproj
+++ b/src/FluentValidation.AspNetCore/FluentValidation.AspNetCore.csproj
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp2.1;netcoreapp3.1</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;netcoreapp2.1;netcoreapp3.1</TargetFrameworks>
     <AssemblyName>FluentValidation.AspNetCore</AssemblyName>
     <PackageId>FluentValidation.AspNetCore</PackageId>
     <Product>FluentValidation.AspNetCore</Product>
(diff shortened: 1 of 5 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/FluentValidation/FluentValidation?pullRequest=1505) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/FluentValidation/FluentValidation?pullRequest=1505) before we can accept your contribution.<br/><br/>:x: dhwed [sign now](https://cla.dotnetfoundation.org/FluentValidation/FluentValidation?pullRequest=1505)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/FluentValidation/FluentValidation?pullRequest=1505) it.</sub>

JeremySkinner:
Hello, we used to support this scenario in FluentValidation 8, and a conscious decision was made to remove this for for FluentValidation 9. 

With FluentValidation 9, we only support asp.net core running on .net core 2.1 or 3.1, and we will not be re-adding support for asp.net core running on .net framework, or older versions of .net core. 

If you need support for this, you will need to stick with FluentValidation 8.6.2. Once you migrate to a modern version of .net, you’ll be able to upgrade to FluentValidation 9. 

I’m sorry if this is disappointing, but it isn’t practical for us to keep supporting this scenario (or old versions of .net core, which we also dropped with the 9.0 release).

You will need to use FluentValidation 8.x if you need support for this.

dhwed:
Thanks for the quick response and the explanation.

Would you be willing to accept a PR to just change the `netcoreapp2.1` dependency from `Microsoft.AspNetCore.Mvc` to `Microsoft.AspNetCore.Mvc.ViewFeatures`?  If so, can it also be backported to make an upstream 8.6.3 version targetting `netstandard2.0` with the reduced dependency?

If not, we can always package up our own version on our internal nuget repo until we upgraded to `netcoreapp3.1` or later and then this doesn't matter anymore.

JeremySkinner:
Changing the dependency graph like this is technically a breaking change which makes me a little hesitant, but in reality I can't foresee it causing issues, so yes if you'd like to send me 2 PRs - one for the master branch and one for the 8.x branch - then I'll do some testing, and if it all looks OK I'll merge them.

PR #1509, opened 2020-08-28
remove razor dependency for 8.x

dhwed:
The razor dependencies in `Microsoft.AspNetCore.Mvc` bring in several additional megabytes of unused binaries to each of our applications. This library does not require any of the features for cshtml file compilation, so we can change the dependency to `Microsoft.AspNetCore.ViewFeatures`.


JeremySkinner:
Don't worry about the build failure for this - it's due to to a missing .net core 2.2 SDK on the build runner. I'll fix that.

JeremySkinner:
I've pushed out 8.6.3 with this change

dhwed:
Awesome - thanks for getting this released so quickly.

My bin directories are now all 12 MiB smaller.

PR #1545, opened 2020-10-10
Razor pages client ruleset

DavidJohnWilliams:
Hi Jeremy,

I have created a PR for the new `RuleSetForClientSideMessagesPageFilter`.

I have realised that because my previous branch was tracking net5 rather than its own remote that my first commit went on the net5 branch. So having taken a second branch that commit is also now included with this PR.

I am hoping you will be able to cherry pick the 3 later commits from this branch... but honestly I am still getting up to speed with git so apologies.

If you need me to create a new PR or anything I can do.

Best regards

Dave

### DIFF
--- a/Changelog.txt
+++ b/Changelog.txt
@@ -2,6 +2,7 @@
 Support for .NET 5
 Improvements to LanguageManager's lazy loading of resources.
 Deprecate IStringSource and its implementors. Use delegates instead.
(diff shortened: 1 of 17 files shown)


JeremySkinner:
Thanks, I’ll try and review this soon but it would be great if you can rebase it not to include the commits from the other PR. You should be able to create your branch based on the state of the net5 branch from before the other commits. If you’re not able to then I’ll sort it out later but it’ll take me longer to get round to it.

Issue #1544, opened 2020-10-09
RuleSetForClientSideMessagesAttribute for Razor Pages PageModel


DavidJohnWilliams:
Hi Jeremy,

I have a potential resolution for this one as well but I thought it best to open some discussion first.

The current `RuleSetForClientSideMessagesAttribute` cannot be used with Razor Pages as it is an ActionFilter.

### Option 1

Use a [PageFilter](https://docs.microsoft.com/en-us/aspnet/core/razor-pages/filter?view=aspnetcore-3.1) which is added for all pages. This then checks for the presence of the `RuleSetForClientSideMessagesAttribute` and sets the value in `HttpContext` accordingly.

I have tested this it works fine. Quite neat actually and requires no changes to the current attribute. One possible issue is that it will call the filter for all pages unless the developer specifies which pages to run it for.

It could possibly be added in `FluentValidationMvcExtensions` via configuration.

### Option 2

Add a new attribute e.g. `PageModelRuleSetForClientSideMessagesAttribute` which inherits `ResultFilterAttribute` and does the same thing as the `RuleSetForClientSideMessagesAttribute`.

This also passes tests and doesn't require any addition config. But of course requires a different attribute and can only be applied for the whole page and all handlers.

### Conclusion

I think it might be worth implementing both and updating the documentation accordingly. If you would prefer only one or the other let me know and I will adjust before submitting the PR.

Also are you happy for me to to the docs update this time to save you having to do it?

And are you happy if I add some configuration options to the `FluentValidationMvcExtensions` to allow adding the page filter or would you prefer to just leave it to developers to add manually and just document it?

Best regards

Dave

JeremySkinner:
Hi Dave, I think I prefer option 1, so let’s go for that if you’d like to implement it. Let’s
enable it by default (in the RegisterServices method I think) - I’m not sure whether it really needs a config option to disable it, but let’s add one anyway.

JeremySkinner:
The .net5 branch has been merged into master, and this will be included in the 9.3 release.

PR #1546, opened 2020-10-10
Razor pages client ruleset

DavidJohnWilliams:
Hi Jeremy,

Apologies again for messing up the branching. I tried a whole load of things and I don't really know what is going on.

In the end I merged the upstream branch according to github's documentation. And I created a new branch from net5 and cherry picked my commits into it.

For some reason it still showing the 4 previous commits. Although the actual changes are now showing as only the changes I made in this branch.

I have no idea what went wrong but hopefully it can be worked out?

Best regards

Dave

### DIFF
--- a/src/FluentValidation.AspNetCore/RuleSetForClientSideMessagesAttribute.cs
+++ b/src/FluentValidation.AspNetCore/RuleSetForClientSideMessagesAttribute.cs
@@ -34,7 +34,7 @@ namespace FluentValidation.AspNetCore {
 		/// By default, only rules not in a ruleset will be used.
 		/// </summary>
 		/// <param name="context"></param>
-		public void SetRulesetForClientValidation(HttpContext context) {
+		internal void SetRulesetForClientValidation(HttpContext context) {
 			SetRulesetForClientValidation(context, _ruleSets);
 		}
 
--- a/src/FluentValidation.AspNetCore/RuleSetForClientSideMessagesAttribute.cs
+++ b/src/FluentValidation.AspNetCore/RuleSetForClientSideMessagesAttribute.cs
@@ -1,7 +1,7 @@
 namespace FluentValidation.AspNetCore {
 	using System;
-  using FluentValidation.Internal;
-  using Microsoft.AspNetCore.Http;
+	using FluentValidation.Internal;
+	using Microsoft.AspNetCore.Http;
 	using Microsoft.AspNetCore.Mvc.Filters;
 
 	/// <summary>
@@ -16,20 +16,28 @@ namespace FluentValidation.AspNetCore {
 		}
 
 		public RuleSetForClientSideMessagesAttribute(params string[] ruleSets) {
-			this._ruleSets = ruleSets;
+			_ruleSets = ruleSets;
 		}
 
 		public override void OnActionExecuting(ActionExecutingContext filterContext) {
 			var contextAccessor = filterContext.HttpContext.RequestServices.GetService(typeof(IHttpContextAccessor));
 
-			if(contextAccessor == null) {
+			if (contextAccessor == null) {
 				throw new InvalidOperationException("Cannot use the RuleSetForClientSideMessagesAttribute unless the IHttpContextAccessor is registered with the service provider. Make sure the provider is registered by calling services.AddSingleton<IHttpContextAccessor, HttpContextAccessor>(); in your Startup class's ConfigureServices method");
 			}
 
 			SetRulesetForClientValidation(filterContext.HttpContext, _ruleSets);
 		}
 
-		
+		/// <summary>
+		/// Allows the ruleset used for generating clientside metadata to be overriden.
+		/// By default, only rules not in a ruleset will be used.
+		/// </summary>
+		/// <param name="context"></param>
+		public void SetRulesetForClientValidation(HttpContext context) {
+			SetRulesetForClientValidation(context, _ruleSets);
+		}
+
 		/// <summary>
 		/// Allows the ruleset used for generating clientside metadata to be overriden.
 		/// By default, only rules not in a ruleset will be used.
@@ -53,7 +61,7 @@ namespace FluentValidation.AspNetCore {
 				return context?.Items[_key] as string[] ?? new[] { RulesetValidatorSelector.DefaultRuleSetName };
 
 			}
-			return new[] { RulesetValidatorSelector.DefaultRuleSetName};
+			return new[] { RulesetValidatorSelector.DefaultRuleSetName };
 		}
 	}
 }
--- a/src/FluentValidation.AspNetCore/RuleSetForClientSideMessagesPageFilter.cs
+++ b/src/FluentValidation.AspNetCore/RuleSetForClientSideMessagesPageFilter.cs
@@ -26,7 +26,7 @@ namespace FluentValidation.AspNetCore {
 	/// <summary>
 	/// Determines which ruleset should be used when deciding which validators should be used to generate client-side messages.
 	/// </summary>
-	public class RuleSetForClientSideMessagesPageFilter : IAsyncPageFilter {
+	internal class RuleSetForClientSideMessagesPageFilter : IAsyncPageFilter {
 
 		public Task OnPageHandlerSelectionAsync(PageHandlerSelectedContext context) {
 			return Task.CompletedTask;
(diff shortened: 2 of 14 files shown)


JeremySkinner:
That's ok, I think you probably merged upstream back into your repo rather than rebasing or resetting the history, which is why it's showing the commits again as a diverged history. I've rewritten your branch history so it only contains the new commits, and I'll try and give this a proper review in a couple days. Thanks again!

src/FluentValidation.AspNetCore/RuleSetForClientSideMessagesAttribute.cs
```diff
@@ -16,20 +16,28 @@ public class RuleSetForClientSideMessagesAttribute : ActionFilterAttribute {
...
 			SetRulesetForClientValidation(filterContext.HttpContext, _ruleSets);
 		}
 
-		
+		/// <summary>
+		/// Allows the ruleset used for generating clientside metadata to be overriden.
+		/// By default, only rules not in a ruleset will be used.
+		/// </summary>
+		/// <param name="context"></param>
+		public void SetRulesetForClientValidation(HttpContext context) {
```
JeremySkinner, line 37: Let's make this internal

JeremySkinner:
Looks good to me, I made a couple of minor tweaks:

- Some minor formatting changes
- Made the new class/method internal (no need to include it in the public api surface)

I'm still unsure whether we need a configuration option to enable/disable this, I'm tempted to remove the option and just always have it enabled - I can't really think of any reason why having this always enabled would cause an issue. I'll have a think about it over the next couple of days and then get it merged. Thanks again for doing this.

src/FluentValidation.AspNetCore/RuleSetForClientSideMessagesPageFilter.cs
```diff
@@ -0,0 +1,47 @@
...
+
+namespace FluentValidation.AspNetCore {
+	using System.Linq;
+	using System.Threading.Tasks;
+	using Microsoft.AspNetCore.Mvc.Filters;
+
+	/// <summary>
+	/// Determines which ruleset should be used when deciding which validators should be used to generate client-side messages.
+	/// </summary>
+	internal class RuleSetForClientSideMessagesPageFilter : IAsyncPageFilter {
```
JeremySkinner, line 29: Changed this to internal too

DavidJohnWilliams: My only comment about this is that it would prevent people from adding it themselves to a specific page. The point of the configuration option was to allow removing the filter from all requests and adding it to only specific pages. If we aren't going with the option in the end they the filter can be internal for sure.

JeremySkinner: Yeah let's remove the configuration option and leave the classes internal. If someone asks for this functionality then we can reconsider it in the future, but I prefer not to make the public api surface larger than necessary unless there's a good reason (as this increases my support burden).

DavidJohnWilliams: I am happy with that :)

JeremySkinner:
One quick thought about this (you may have considered it already), is does this actually need to be at the method level? Is there a use case for wanting a different ruleset applied for GET vs POST, or would it make more sense to apply this at the class level like a regular filter attribute? 

If so, could we then use the attribute class directly rather than wrapping it in a page filter?

DavidJohnWilliams:
> Is there a use case for wanting a different ruleset applied for GET vs POST

I don't see a particular use case for different validation between `GET` and `POST` as such. But definitely between `Get/PostStep1` and `Get/PostStep2` as an example. I have see this sort of thing used (albeit rarely) when using the same page so do an initial add of an object and later a more comprehensive edit.

> If so, could we then use the attribute class directly rather than wrapping it in a page filter?

That was essentially option 2 from issue #1544. And actually after looking a bit further I can see that [`ActionFilterAttribute`](https://github.com/aspnet/Mvc/blob/master/src/Microsoft.AspNetCore.Mvc.Core/Filters/ActionFilterAttribute.cs) actually implements `IResultFilter` so you can simply override `OnResultExecuting` and do the same thing as `OnActionExecuting`.

That would allow using the same attribute with no other changes to the code, but means that the same validation will be applied for all handlers.

Like I said before, I think that for the majority of use cases a `PageModel` attribute would be sufficient. The use for the page filter is perhaps more of an edge case. But I can still see a definite use for it.

My suggestion would be:

1. Override `OnResultExecuting` so the attribute can be used with page model and document that.
2. Choose either to include the page filter and put back the switch and default it to `false`. OR just add documentation which shows a way that you can implement page filter yourself but provide the necessary publicly accessible `SetRulesetForClientValidation` on the attribute.

One note on the `SetRulesetForClientValidation` is that I had considered moving it to a utility class when I though it would be necessary to make separate Attributes. But as it simply needs to get and set a key in the `context.Items` perhaps a better solution would be to have an extension method for both `Get` and `Set` something like:

```
    public static HttpContextExtensions
    {
        private const string _key = "_FV_ClientSideRuleSet";

        /// <summary>
        /// Allows the ruleset used for generating clientside metadata to be overriden.
        /// By default, only rules not in a ruleset will be used.
        /// </summary>
        /// <param name="context"></param>
        /// <param name="ruleSets"></param>
        public static void SetRulesetForClientValidation(this HttpContext context, string[] ruleSets) {
            context.Items[_key] = ruleSets;
        }

        /// <summary>
        /// Gets the rulesets used to generate clientside validation metadata.
        /// </summary>
        /// <param name="context"></param>
        /// <returns></returns>
        public static string[] GetRuleSetsForClientValidation(this HttpContext context) {
            // If the httpContext is null (for example, if IHttpContextProvider hasn't been registered) then just assume default ruleset.
            // This is OK because if we're actually using the attribute, the OnActionExecuting will have caught the fact that the provider is not registered. 

            if (context?.Items != null && context.Items.ContainsKey(_key)) {
                return context?.Items[_key] as string[] ?? new[] { RulesetValidatorSelector.DefaultRuleSetName };
            }
           
            return new[] { RulesetValidatorSelector.DefaultRuleSetName };
        }
    }
```
Then add `public IReadOnlyCollection<string> RuleSets => Array.AsReadOnly(_ruleSets);` to the `RuleSetForClientSideMessagesAttribute` so that the rulesets can be read out for use in the PageFiler.

That way FluentValidation doesn't have to implement the page filter. But there is nothing stopping others from doing do.

I'm not sure what the best way to demo this would be. I can make another branch (properly this time) and make the changes on there and either submit an second PR or you can just look at the commits in that branch?

JeremySkinner:
> That would allow using the same attribute with no other changes to the code, but means that the same validation will be applied for all handlers.

Yes I think I prefer that. Initially I preferred option 1, as I didn't like the idea of having 2 attributes, but actually if they can both use the same attribute I'd prefer not to implement the separate filter class and just let the attribute handle it as a normal filter, which can be applied to the page model, and then there's no need for the global filter class. 

> Override OnResultExecuting so the attribute can be used with page model and document that.

Yes sounds good.

> Choose either to include the page filter and put back the switch and default it to false. OR just add documentation which shows a way that you can implement page filter yourself but provide the necessary publicly accessible SetRulesetForClientValidation on the attribute.

Let's omit the page filter.

> perhaps a better solution would be to have an extension method for both Get and Set something like:

We already have an extension method for set (`SetRulesetForClientsideMessages` in the `ValidationResultExtension` class), but it takes a ControllerContext as the argument, so we'd need to add a second overload that takes the appropriate context class for razor pages too. For backwards compatibility these will need to stay in the `ValidationResultExtension` class for now (as moving extension methods to another class is a breaking change from the runtime's point of view), but we can move that into a separate class in a future major version.

> Then add public IReadOnlyCollection<string> RuleSets => Array.AsReadOnly(_ruleSets); to the RuleSetForClientSideMessagesAttribute

I'm happy with just exposing a normal getter for the array. No need to make it truly readonly.

> I'm not sure what the best way to demo this would be. I can make another branch (properly this time) and make the changes on there and either submit an second PR or you can just look at the commits in that branch?

Another branch/PR would be good, then I can merge that one and close this one (give me a shout if you get stuck with creating the branch)

Thanks!

DavidJohnWilliams:
Just prepping the new PR. I reset my net5 branch to be the same as the upstream net5 branch (without all the extra commits and merge etc.) and it seems to have resolved the problem with creating new branches off of my forked net5 branch.

One question: Do you have anything against expression bodied methods/constructors etc?

VS is giving me messages about it and I would personally prefer to make it so, but happy to follow whatever you would like.

JeremySkinner:
Very happy to use them for new stuff, but don’t really see a need to retroactively change existing code

DavidJohnWilliams:
Closing as this was replaced by #1548

PR #1548, opened 2020-10-13
Razor pages client ruleset attribute

DavidJohnWilliams:
Hi Jeremy,

This is the updated PR for #1544 

I have centralized the logic for the `SetRulesetForClientsideMessages` extension which neatly resolves the issue of using it with page handlers anyway.

So you can use the attribute to apply it for all handlers on the `PageModel` or you can specify the ruleset via the extension on a specific handler.

One thing that might be worth mentioning in the docs is that the attribute takes precedence over the extension method. Add will overwrite the ruleset specified in the handler. I suspect the same applies to a controller action (not tested).

Good job you had that one question ;p

### DIFF
--- a/src/FluentValidation.AspNetCore/RuleSetForClientSideMessagesAttribute.cs
+++ b/src/FluentValidation.AspNetCore/RuleSetForClientSideMessagesAttribute.cs
@@ -1,6 +1,5 @@
 namespace FluentValidation.AspNetCore {
 	using System;
-	using FluentValidation.Internal;
 	using Microsoft.AspNetCore.Http;
 	using Microsoft.AspNetCore.Mvc.Filters;
 
@@ -8,7 +7,7 @@ namespace FluentValidation.AspNetCore {
 	/// Specifies which ruleset should be used when deciding which validators should be used to generate client-side messages.
 	/// </summary>
 	public class RuleSetForClientSideMessagesAttribute : ActionFilterAttribute {
-		private const string _key = "_FV_ClientSideRuleSet";
+
 		private readonly string[] _ruleSets;
 
 		public RuleSetForClientSideMessagesAttribute(string ruleSet) {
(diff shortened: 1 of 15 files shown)


src/FluentValidation.AspNetCore/RuleSetForClientSideMessagesAttribute.cs
```diff
@@ -1,59 +1,46 @@
...
+		public RuleSetForClientSideMessagesAttribute(string ruleSet) => _ruleSets = new[] { ruleSet };
 
-		public override void OnActionExecuting(ActionExecutingContext filterContext) {
-			var contextAccessor = filterContext.HttpContext.RequestServices.GetService(typeof(IHttpContextAccessor));
+		public RuleSetForClientSideMessagesAttribute(params string[] ruleSets) => _ruleSets = ruleSets;
 
-			if(contextAccessor == null) {
-				throw new InvalidOperationException("Cannot use the RuleSetForClientSideMessagesAttribute unless the IHttpContextAccessor is registered with the service provider. Make sure the provider is registered by calling services.AddSingleton<IHttpContextAccessor, HttpContextAccessor>(); in your Startup class's ConfigureServices method");
-			}
+		public override void OnActionExecuting(ActionExecutingContext filterContext) => SetRulesetOnExecuting(filterContext);
```
JeremySkinner, line 17: Do we need `OnActionExecuting` at all anymore, now that we have `OnResultExecuting` overridden? I think having both will mean `SetRulesetOnExecuting` will end up running twice for MVC actions.

JeremySkinner:
One question about the filter but apart from that it looks good to me!

src/FluentValidation.AspNetCore/RuleSetForClientSideMessagesAttribute.cs
DavidJohnWilliams: You are correct. Not needed so I have pushed a commit which removes `OnActionExecuting`

DavidJohnWilliams:
Do you want me to add anything to documentation about the precedence of the attribute over the extension method?

JeremySkinner:
No I think that's fine for now - I can't imagine a scenario where someone would use both. Can always add it later if needed. Thanks again for doing this! 

I'll squash the commits into 1 and then merge this to the net5 branch,

PR #1562, opened 2020-10-28
Remove support for .NET 4.x, netstandard2.0 and netcoreapp2.1

JeremySkinner:
Going forward we'll only support the latest LTS release (currently netcoreapp3.1) plus any still-supported non-LTS release (net5)

### DIFF
--- a/global.json
+++ b/global.json
@@ -3,5 +3,5 @@
       "version": "5.0.100",
       "rollForward": "latestFeature"
     },
-    "others": ["3.1.201", "2.2.105"]
+    "others": ["3.1.201"]
 }
--- a/src/FluentValidation.AspNetCore/FluentValidation.AspNetCore.csproj
+++ b/src/FluentValidation.AspNetCore/FluentValidation.AspNetCore.csproj
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp2.1;netcoreapp3.1;net5.0</TargetFrameworks>
+    <TargetFrameworks>netcoreapp3.1;net5.0</TargetFrameworks>
     <AssemblyName>FluentValidation.AspNetCore</AssemblyName>
     <PackageId>FluentValidation.AspNetCore</PackageId>
     <Product>FluentValidation.AspNetCore</Product>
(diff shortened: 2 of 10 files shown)


augustoproiete:
@JeremySkinner Is there a strong reason to remove `netstandard2.0` from core libraries (i.e. `FluentValidation`, `FluentValidation.DependencyInjectionExtensions`) instead of keeping it and adding `netstandard2.1` as an additional target?

It would be nice to allow apps that aren't ready for .NET Standard 2.1 (e.g. UWP, Unity, ...) to benefit from v10 features, if it's not a lot of work on your end.

That would also allow .NET Framework 4.6.1+ apps to benefit from it as well (even if .NET Framework is not officially supported).

JeremySkinner:
@augustoproiete Supporting old frameworks indefinitely prevents being able to evolve the library. For example, many of the newer c# 8 and c# 9 features are only available on .net core 3.1 / netstandard2.1 or newer. Additionally, the more frameworks we support the more of a support, maintenance, and testing burden this adds for me (I am just 1 person that does this in my spare time).

Microsoft are removing netstandard2.0 support from many of their own libraries going forward (for example, newer versions of EF core now only support netstandard2.1 at minimum), so this seems like a good time to drop legacy frameworks.  .NET 4.x is definitely something that I no longer wish to support.

This change will only go in for FluentValidation 10. FluentValidation 9 will still be available for those who need to target netstandard2.0, netcoreapp2.1 or net461+

augustoproiete:
@JeremySkinner Sure, I understand, and didn't imply .NET Standard 2.0 support should continue indefinitely... I was just wondering which features of .NET Standard 2.1 the FluentValidation 10 release will effectively be using (if any) that would cause any burden to maintain .NET Standard 2.0 support for a little longer until it's technically justifiable to drop it - you would know better.

ps: I think EF Core dropping `netstandard2.0` is not as harsh as FluentValidation doing it. In my mind, EF Core is meant to be used on the server where it's generally much easier to upgrade to newer versions of .NET compared to Windows client apps. FluentValidation is widely used in all kinds of apps, including Windows client apps, and the poor souls stuck on .NET 4.8 WPF/WinForms land will also be stuck in FluentValidation 9.x :upside_down_face:

JeremySkinner:
> I was just wondering which features of .NET Standard 2.1 the FluentValidation 10 release will effectively be using (if any)

The focus of the 10.0 release is to remove deprecated code in the core library and drop old framework support- there will be very little in the way of new features. This an incremental shift which will open the doors to use newer language features in future versions. Currently FV can't make use of tuples (because net46 can't consume System.ValueTuple properly). I also want to make use of default interface methods to be able to simplify how the fluent interface is built, as well as using async enumerables in FV's async support and C#9 records to simplify some type definitions. 

> and the poor souls stuck on .NET 4.8 WPF/WinForms land will also be stuck on FluentValidation 9.x

You say that like it's a bad thing. If you're using FV 9 on net4x, there really is no need to upgrade to 10 when it's out. As I said before, 9.x will continue to be available.

PR #1586, opened 2020-11-18
Add ImplicitlyValidateRootCollectionElements MVC option

### LINKED ISSUE #1585: Implicit validation of root collection models
**Is your feature request related to a problem? Please describe.**
I would like to be able to validate the elements of a root model that is a collection type without enabling implicit validation of child properties or having to define an explicit collection validator.

*Example model and controller action method:*
```c#
public class Model { 
    public string Name { get; set; }
}

public class ModelValidator : AbstractValidator<Model> {
    public ModelValidator() => RuleFor(m => m.Name).NotNull().WithMessage("NotNull");
}

[ApiController]
public class TestController : ControllerBase {
    [HttpPost]
    public IActionResult ControllerValidation(IEnumerable<Model> model) => Ok();
}
```

Given the model and controller above and the equivalent rule implemented using a `[Required("NotNull")]` data model annotation, the errors returned by the default model validation framework would be of the following format.

*Example (partial) error response:*
```json
{ "[0].Name": ["NotNull"] }
```

While it is possible to produce the same response by enabling implicit validation of child properties, it obviously has the side effect that any child properties of `Model` will also be automatically validated – however, I'd like to control this validation explicitly.

I can define an explicit collection validator like so:

```c#
public class ListOfModelValidator : AbstractValidator<List<Model>> {
    public ListOfModelValidator() => RuleForEach(model => model).SetValidator(new ModelValidator());
}
```

However, this has the side effect that the response will now be as below, where the lambda parameter name is used as the property name.

```json
{ "model[0].Name": ["NotNull"] }
```

Of course, `WithName` can be used to override the property name but not to set it to a null or an empty string. I think this behaviour makes sense when using `RuleForEach` for a child property but I'm not sure it does when it is the root model that is the collection. Equally, while `[0].Name` may not seem like an ideal naming convention, it matches the ASP.NET Core default.

**Describe the solution you'd like**
One option might be to introduce a `ImplicitlyValidateRootCollectionElements` property to `FluentValidationMvcConfiguration`. It would default to `false` but when set to `true` validation of collection elements would occur the same way it does when `ImplicitlyValidateChildProperties` is set to `true`. If both `ImplicitlyValidateRootCollectionElements` and `ImplicitlyValidateChildProperties` are set to `true` then `ImplicitlyValidateRootCollectionElements` would simply be ignored and the behaviour of `ImplicitlyValidateChildProperties` would be as it is now.

**Describe alternatives you've considered**
1. Rewrite the property name after validation by overriding `Validate` or by implementing `IValidatorInterceptor`. However, using either of those options to rewrite the property name after validation just feels wrong.
2. Redesign the model and/or response (by using `WithName` perhaps). However, this may not be an option where an API is public and/or already in production.

**Additional context**
This may be a bit of an edge case but I think it may be useful for developers inheriting an already published endpoint and/or converting annotation-based validation to FluentValidation (as is my case).

I'd be happy to contribute a PR (either for what I've suggested above or an alternative) if the suggestion fits with the existing design and roadmap.

notnotchris:
Fixes #1585

### DIFF
--- a/docs/aspnet.md
+++ b/docs/aspnet.md
@@ -166,7 +166,7 @@ By default, you must create a specific collection validator or enable implicit c
 public ActionResult DoSomething(List<Person> people) => Ok();
 ```
 
-With implicit child property validation enabled, each person element will be validated but so will any child properties of each `Person`. When running an ASP.NET MVC application, you can also optionally enable implicit validation for root collection elements only. For example, if you want each `Person` element to be validated but not its child properties. This can be done by setting `ImplicitlyValidateRootCollectionElements` to true:
+With implicit child property validation enabled (see above), you don't have to explicitly create a collection validator class as each person element in the collection will be validated automatically. However, any child properties on the `Person` object will be automatically validated too meaning you can no longer use `SetValidator`. If you don't want this behaviour, you can also optionally enable implicit validation for root collection elements only. For example, if you want each `Person` element in the collection to be validated automatically, but not its child properties you can set `ImplicitlyValidateRootCollectionElements` to true:
 
 ```csharp
 services.AddMvc().AddFluentValidation(fv => {
--- a/docs/aspnet.md
+++ b/docs/aspnet.md
@@ -160,10 +160,10 @@ Note that if you enable this behaviour you should not use `SetValidator` for chi
 
 ### Implicit Validation of Collection-Type Models
 
-By default, you must create a specific collection validator or enable implicit child property validation to validate a model that is of a collection type. For example, no validation of the following model will occur with the default settings.
+By default, you must create a specific collection validator or enable implicit child property validation to validate a model that is of a collection type. For example, no validation of the following model will occur with the default settings unless you define a validator that inherits from `AbstractValidator<List<Person>>`.
 
 ```csharp
-public ActionResult DoSomething(ICollection<Person> people) => Ok();
+public ActionResult DoSomething(List<Person> people) => Ok();
 ```
 
 With implicit child property validation enabled, each person element will be validated but so will any child properties of each `Person`. When running an ASP.NET MVC application, you can also optionally enable implicit validation for root collection elements only. For example, if you want each `Person` element to be validated but not its child properties. This can be done by setting `ImplicitlyValidateRootCollectionElements` to true:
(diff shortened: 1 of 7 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/FluentValidation/FluentValidation?pullRequest=1586) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/FluentValidation/FluentValidation?pullRequest=1586) before we can accept your contribution.<br/><br/>:x: notnotchris [sign now](https://cla.dotnetfoundation.org/FluentValidation/FluentValidation?pullRequest=1586)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/FluentValidation/FluentValidation?pullRequest=1586) it.</sub>

JeremySkinner:
Thanks - I'll try and give this a review over the weekend

JeremySkinner:
Due to various circumstances I haven’t had a chance to dive into this yet. It’s still on my radar but probably won’t be something I get to until next year.

notnotchris:
> Due to various circumstances I haven’t had a chance to dive into this yet. It’s still on my radar but probably won’t be something I get to until next year.

No problem. Thanks for the update.

docs/aspnet.md
```diff
@@ -158,6 +158,24 @@ services.AddMvc().AddFluentValidation(fv => {
 
 Note that if you enable this behaviour you should not use `SetValidator` for child properties, or the validator will be executed twice.
 
+### Implicit Validation of Collection-Type Models
+
+By default, you must create a specific collection validator or enable implicit child property validation to validate a model that is of a collection type. For example, no validation of the following model will occur with the default settings unless you define a validator that inherits from `AbstractValidator<List<Person>>`.
+
+```csharp
+public ActionResult DoSomething(List<Person> people) => Ok();
```
JeremySkinner, line 166: I changed this example from `ICollection` to `List` for clarity, as by default validating with interface types is slightly misleading. When you specify an interface collection type as a controller parameter, MVC will internally create a `List<T>`, and as such the validator must be defined as `AbstractValidator<List<T>>` not `AbstractValidator<ICollection<T>>`. It's always better to be explicit with controller parameters, as using interface collection types here 'hides' the magic.

Issue #1617, opened 2021-01-20
Add Information in the RuleSets documentation page


lucasgit2000:
**Is your feature request related to a problem? Please describe.**
I have to look somewhere else outside the doc page a line of code that was needed to implement RuleSets

**Describe the solution you'd like**
It would be great if in the RuleSets documentation we could also find a piece of information that describes the only way to use options method overload is if we put ```using FluentValidation``` in our code, otherwise it won't work.

**Describe alternatives you've considered**
Maybe the extension or the method with options overload also be located in the same namespace as ```using FluentValidation.Results```?

**Additional context**
```
using FluentValidation

var result = await validator.ValidateAsync(req, options: options => {
options.IncludeAllRuleSets();
});
```

Forgive me if I'm missing something, just found this library and saw the great potential on it, I just wanna help.

JeremySkinner:
Thanks for the suggestion. Many of the methods in FluentValidation are extension methods and require the FluentValidation namespace to imported via a using statement. 

This particular method will likely become an instance method in the 10.0 release.

lucasgit2000:
#1618 opened, I hope you'll attend, in case I have to do something else, I'm all ears.

PR #1618, opened 2021-01-21
FluentValidation namespace explanation in RuleSet doc page

lucasgit2000:
The issue for more details: https://github.com/FluentValidation/FluentValidation/issues/1617#issuecomment-763958776

I put a note tag below of the first example so anyone can realize they have to declare the namespace to the method with ```options``` overload to work.


JeremySkinner:
Merged, thanks!

Issue #1625, opened 2021-01-27
'NotEmpty' validation throws 'InvalidOperationException' on 'ImmutableArray<T>' property


julealgon:
**Which version of FluentValidation are you using?**
9.3.0

**Which version of ASP.NET are you using?**
Non-ASP.NET related use-case

**Describe the issue that you're having**
I have a property of type `ImmutableArray<byte>` that I want to guarantee is not empty during validation. However, after applying the `NotEmpty` validation on the property and testing without providing the data, an `InvalidOperationException` is thrown:

```
System.InvalidOperationException: This operation cannot be performed on a default instance of ImmutableArray<T>.  Consider initializing the array, or checking the ImmutableArray<T>.IsDefault property.
   at System.Collections.Immutable.ImmutableArray`1.ThrowInvalidOperationIfNotInitialized()
   at System.Collections.Immutable.ImmutableArray`1.System.Collections.ICollection.get_Count()
   at FluentValidation.Validators.NotEmptyValidator.IsValid(PropertyValidatorContext context) in C:\Projects\FluentValidation\src\FluentValidation\Validators\NotEmptyValidator.cs:line 38
   at FluentValidation.Validators.PropertyValidator.Validate(PropertyValidatorContext context) in C:\Projects\FluentValidation\src\FluentValidation\Validators\PropertyValidator.cs:line 76
   at FluentValidation.Internal.PropertyRule.InvokePropertyValidator(IValidationContext context, IPropertyValidator validator, String propertyName) in C:\Projects\FluentValidation\src\FluentValidation\Internal\PropertyRule.cs:line 480
   at FluentValidation.Internal.PropertyRule.ValidateAsync(IValidationContext context, CancellationToken cancellation) in C:\Projects\FluentValidation\src\FluentValidation\Internal\PropertyRule.cs:line 401
   at FluentValidation.AbstractValidator`1.ValidateAsync(ValidationContext`1 context, CancellationToken cancellation) in C:\Projects\FluentValidation\src\FluentValidation\AbstractValidator.cs:line 139
```

Here is a sample model class:

```csharp
    public record UploadDocumentRequest(
        string? FileName,
        ImmutableArray<byte> FileContents);
```

And here is the associated validator:

```csharp
    public class UploadDocumentRequestValidator : AbstractValidator<UploadDocumentRequest>
    {
        public UploadDocumentRequestValidator()
        {
            _ = this.RuleFor(r => r.FileContents).NotEmpty();
[...]

PR #1626, opened 2021-01-27
Ensure Empty/NotEmpty work with ImmutableArray.

JeremySkinner:
For #1625

### DIFF
--- a/src/FluentValidation/Validators/EmptyValidator.cs
+++ b/src/FluentValidation/Validators/EmptyValidator.cs
@@ -19,10 +19,9 @@
 #endregion
 
 namespace FluentValidation.Validators {
-	using System;
-	using System.Collections;
-	using Resources;
-	using System.Linq;
+#if !NET461 && !NETSTANDARD2_0
+	using System.Collections.Immutable;
+#endif
 
 	public class EmptyValidator : PropertyValidator, IEmptyValidator {
 		readonly object _defaultValueForType;
@@ -39,6 +38,20 @@ namespace FluentValidation.Validators {
 			return Localized(nameof(EmptyValidator));
 		}
 	}
+
+	#if !NET461 && !NETSTANDARD2_0
+	internal class EmptyImmutableArrayValidator<TElement> : PropertyValidator, INotEmptyValidator {
+		protected override bool IsValid(PropertyValidatorContext context) {
+			if (context.PropertyValue == null) return true;
+			var value = (ImmutableArray<TElement>)context.PropertyValue;
+			return value.IsDefaultOrEmpty;
+		}
+
+		protected override string GetDefaultMessageTemplate() {
+			return Localized(nameof(EmptyValidator));
+		}
+	}
+	#endif
 
 	public interface IEmptyValidator : IPropertyValidator {
 	}
--- a/src/FluentValidation/Validators/EmptyValidator.cs
+++ b/src/FluentValidation/Validators/EmptyValidator.cs
@@ -32,20 +32,7 @@ namespace FluentValidation.Validators {
 		}
 
 		protected override bool IsValid(PropertyValidatorContext context) {
-			switch (context.PropertyValue) {
-				case null:
-				case string s when string.IsNullOrWhiteSpace(s):
-				case ICollection c when c.Count == 0:
-				case Array a when a.Length == 0:
-				case IEnumerable e when !e.Cast<object>().Any():
-					return true;
-			}
-
-			if (Equals(context.PropertyValue, _defaultValueForType)) {
-				return true;
-			}
-
-			return false;
+			return NotEmptyValidator.IsEmpty(context.PropertyValue, _defaultValueForType);
 		}
 
 		protected override string GetDefaultMessageTemplate() {
--- a/src/FluentValidation/Validators/NotEmptyValidator.cs
+++ b/src/FluentValidation/Validators/NotEmptyValidator.cs
@@ -37,7 +37,7 @@ namespace FluentValidation.Validators {
 		}
 
 		internal static bool IsEmpty(object value, object defaultValueForType) {
-			// IStucturalComparable must come before ICollection.
+			// IStructuralComparable must come before ICollection.
 			// ImmutableArray<T> throws when checking its Count property if the array is uninitialized.
 			// Checking IStructuralComparable prevents this.
 			switch (value) {
--- a/src/FluentValidation/Validators/NotEmptyValidator.cs
+++ b/src/FluentValidation/Validators/NotEmptyValidator.cs
@@ -21,6 +21,7 @@
 namespace FluentValidation.Validators {
 	using System;
 	using System.Collections;
+	using System.Collections.Generic;
 	using System.Linq;
 	using Resources;
 
@@ -32,20 +33,28 @@ namespace FluentValidation.Validators {
 		}
 
 		protected override bool IsValid(PropertyValidatorContext context) {
-			switch (context.PropertyValue) {
+			return !IsEmpty(context.PropertyValue, _defaultValueForType);
+		}
+
+		internal static bool IsEmpty(object value, object defaultValueForType) {
+			// IStucturalComparable must come before ICollection.
+			// ImmutableArray<T> throws when checking its Count property if the array is uninitialized.
+			// Checking IStructuralComparable prevents this.
+			switch (value) {
 				case null:
 				case string s when string.IsNullOrWhiteSpace(s):
+				case IStructuralComparable sc when sc.CompareTo(defaultValueForType, Comparer.Default) == 0:
 				case ICollection c when c.Count == 0:
 				case Array a when a.Length == 0:
 				case IEnumerable e when !e.Cast<object>().Any():
-					return false;
+					return true;
 			}
 
-			if (Equals(context.PropertyValue, _defaultValueForType)) {
-				return false;
+			if (Equals(value, defaultValueForType)) {
+				return true;
 			}
 
-			return true;
+			return false;
 		}
 
 		protected override string GetDefaultMessageTemplate() {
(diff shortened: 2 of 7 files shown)


src/FluentValidation/Validators/EmptyValidator.cs
```diff
@@ -32,20 +32,7 @@ public class EmptyValidator : PropertyValidator, IEmptyValidator {
...
-				case IEnumerable e when !e.Cast<object>().Any():
-					return true;
-			}
-
-			if (Equals(context.PropertyValue, _defaultValueForType)) {
-				return true;
-			}
-
-			return false;
+			return NotEmptyValidator.IsEmpty(context.PropertyValue, _defaultValueForType);
```
julealgon, line 35: Nice to get rid of the considerable amount of duplication, although I'd recommend extracting the "IsEmpty" logic to an extension method instead. That would allow both validators to tap into the logic without cross-referencing each other which might be seen as a design smell.

src/FluentValidation/Validators/NotEmptyValidator.cs
```diff
@@ -32,20 +33,28 @@ public class NotEmptyValidator : PropertyValidator, INotEmptyValidator {
 		}
 
 		protected override bool IsValid(PropertyValidatorContext context) {
-			switch (context.PropertyValue) {
+			return !IsEmpty(context.PropertyValue, _defaultValueForType);
+		}
+
+		internal static bool IsEmpty(object value, object defaultValueForType) {
+			// IStucturalComparable must come before ICollection.
```
julealgon, line 40: Small typo on "Structural".

src/FluentValidation/Validators/NotEmptyValidator.cs
```diff
@@ -32,20 +33,28 @@ public class NotEmptyValidator : PropertyValidator, INotEmptyValidator {
...
+			// Checking IStructuralComparable prevents this.
+			switch (value) {
 				case null:
 				case string s when string.IsNullOrWhiteSpace(s):
+				case IStructuralComparable sc when sc.CompareTo(defaultValueForType, Comparer.Default) == 0:
 				case ICollection c when c.Count == 0:
 				case Array a when a.Length == 0:
 				case IEnumerable e when !e.Cast<object>().Any():
-					return false;
+					return true;
```
julealgon, line 50: Have you considered using a switch expression with the `when` component moved to the action, instead of returning explicit booleans?

Something like:
```csharp
return value switch 
{
  null => true,
...
  Array a => a.Length == 0,
  IEnumerable e => !e.Cast<object>().Any(),
...
  _ => Equals(value, defaultValueForType),
}

**Review by julealgon — CHANGES_REQUESTED** [review]

File: src/FluentValidation/Validators/NotEmptyValidator.cs (hunk 3 above)
Comment (JeremySkinner) [reply src/FluentValidation/Validators/NotEmptyValidator.cs:50]: Let's keep the discussion focused on the problem at hand. Syntactic sugar can be addressed at a later stage.

**Comment by JeremySkinner** [comment]
Tests fail as the `IStructuralComparable` approach doesn't allow comparison of an initialized to an uninitialized array either. There's also consideration as to whether uninitialized should be considered empty or not. Internally, `ImmutableArray` conisiders uninitialized and empty to be different (hence the exception) so I wonder whether FluentValidation should do the same, and not try and catch this (especially considering the complexity of this).

File: src/FluentValidation/Validators/NotEmptyValidator.cs (hunk 2 above)
Comment (JeremySkinner) [reply src/FluentValidation/Validators/NotEmptyValidator.cs:40]: Fixed

File: src/FluentValidation/Validators/EmptyValidator.cs (hunk 1 above)
Comment (JeremySkinner) [reply src/FluentValidation/Validators/EmptyValidator.cs:35]: It's an internal implementation detail so I don't really think it matters here. Extension methods are nice for the public-facing API, but for the internals I don't think it's really needed.

**Comment by JeremySkinner** [comment]
Tests fail as the `IStructuralComparable` approach doesn't allow comparison of an initialized to an uninitialized array either. There's also consideration as to whether uninitialized should be considered empty or not. Internally, `ImmutableArray` conisiders uninitialized and empty to be different (hence the exception) - it has 3 states: Uninitialized, Initialized empty, and Initialized not empty.

So I wonder whether FluentValidation shouldn't actually try and catch this, because uninitialized is technically neither Empty or NotEmpty according to the design of ImmutableArray. 

As I see it there are a few options:
- Do not attempt to handle uninitialized ImmutableArrays as uninitialized is a separate state to empty/not-empty. Consumers can check IsDefaultOrEmpty explicitly
- Catch exception and consider this to be empty (not nice)
- Use reflection to find and invoke the `IsDefaultOrEmpty` property as it's not exposed non-generically (seems even worse)

Really not sure what the best way forward is here. What would be your gut feeling @julealgon.

**Comment by julealgon** [comment]
I think a 4th option should be considered as well: create strongly typed validators (one for each category) instead of one that handles every case as `object`. That would allow you to check the properties on `ImmutableArray<T>` without any reflection.

Obviously, this approach takes a bit more work as you'd need extension methods for each of those cases in your switch instead of a single extension method on `object`. On the extension method that handles `ImmutableArray<>`, a custom empty validator could be used that knows how that instance should be handled.

While to me it is very clear that multiple specialized versions of `EmptyValidator` are the most robust and clean approach, it would require non-trivial changes to the existing ones.

**Comment by JeremySkinner** [comment]
I'm not opposed to this idea - this was actually what I originally intended, but it has issues. 

The builder works with 2 generic types - `T` (the root object) and `TProperty` (the property).  For example, the current signature for `NotEmpty` looks like this:

```csharp
public static IRuleBuilderOptions<T,TProperty> NotEmpty<T,TProperty>(this IRuleBuilder<T,Tproperty> rule)
```

I think an overload for ImmutableArray would work fine:

```csharp
public static IRuleBuilderOptions<T,ImmutableArray<TElement>> NotEmpty<T,TElement>(this IRuleBuilder<T,ImmutableArray<TProperty>> rule)
```

The problem comes with interface types (`IEnumerable<T>` etc), or actually anything that can be subclassed/implemented, as the type information would be lost. For example:

```csharp
// Specialised overload for IEnumerable.
public static IRuleBuilderOptions<T,IEnumerable<TElement>> NotEmpty<T,TElement>(this IRuleBuilder<T,IEnumerable<TProperty>> rule)
```

This would act upon anything that implements `IEnumerable<T>`, so if the property is of type `List<T>` then the builder would end up returning a less specialised type in the chain. To illustrate:

```csharp
class Foo {
  public List<Bar> Bars {get; set;}
}

RuleFor(x => x.Bars) // TProperty in the chain is currently List<Bar> 
   .NotEmpty() // NotEmpty overload with IEnumerable generic parameter. Return value acts on IEnumerable<Bar>
   .Must(list => list.Count < 10) // Compiler error, as chain is now returning IEnumerable<Bar> instead of List<Bar>
```

This is undesirable, as the type information needs to be preserved throughout the whole chain, and is why I gave up on this approach in the past. 

Maybe the answer is to only add specialised overloads specifically for `ImmutableArray` and not to bother for other types. I do this in a couple of other places (specifically with `Nullable<T>` for LessThan/GreaterThan etc)

julealgon:
> The problem comes with interface types (IEnumerable<T> etc), or actually anything that can be subclassed/implemented, as the type information would be lost. For example:

I believe you could change the method to make it work. Something like this should preserve the output type:

```csharp
        public static IRuleBuilderOptions<T, TSequenceProperty> NotEmpty<T, TSequenceProperty, TElement>(IRuleBuilder<T, TSequenceProperty> rule)
            where TSequenceProperty : IEnumerable<TElement>
        {
            ...
        }
```

This might pose problems for overload resolution though, so you may want to explore other ways of creating these extensions (may use something more akin to what FluentAssertions does).

---

> Maybe the answer is to only add specialised overloads specifically for ImmutableArray and not to bother for other types.

That might be a good compromise indeed considering what the library already exposes today.

JeremySkinner:
That's something I already considered but unfortunately it doesn't support type inference when you use generic constraints in this way, meaning the consumer would have to explicitly specify each of the type parameters, which is a no-go.

julealgon:
> ...but unfortunately it doesn't support type inference when you use generic constraints in this way, meaning the consumer would have to explicitly specify each of the type parameters, which is a no-go.

Ah indeed, since the compiler does not use generic constraint information to deduce the type arguments, it won't be able to tell `TElement` yeah.

I guess the only way to make this cleaner then would be to avoid the extension method approach for defining the constraints and use an actual generic class like FluentAssertions does, then define the extension methods on the generic class. It might be something you want to look into even though it will potentially cause a breaking change to how the API works by introducing a new element in the call chain to store the generic types at the class level (like `.Should()` does).

JeremySkinner:
Yes that's an approach I've considered before but ultimately decided against. The scope of the breaking change is huge, and a lot of people rely on the current interfaces for their custom validators, so it's probably not worth it. That being said, 10.0 is going  to introduce generics in the internal model, which is a step towards being able to do this so it's something that could be considered in the future (11.0 would be the earliest).

JeremySkinner:
Added separate overload that specifically targets ImmutableArray. This brings up a further issue to do with target frameworks as net461 and netstandard2 require an additional package reference to System.Collections.Immutable to be able to use ImmutableArray. I'm hesitant to add this as users often complain about downloading additional unnecessary packages if they're not directly using them. 

I think I'll park this for now, as the workaround is easy to implement anyway.

jeddytier4:
I wouldn't object if it were a separate package that added the immutable extensions.  Something like FluentValidation.Extensions.Immutables, but yes it would be a problem for me to bring in another library for projects that don't use Immutables.

JeremySkinner:
@jeddytier4 That's not something I'm personally interested in developing or supporting, but if you (or someone else) wanted to, then please feel free. 

I'm going to go ahead and close this for the reasons above.

PR #1633, opened 2021-02-02
Add test-reporter CI step

dorny:
This PR:
1. Configures test results filename to `$(MSBuildProjectName)-$(TargetFramework).trx`
1. Adds CI step [test-reporter](https://github.com/dorny/test-reporter) to display report from test results 

Here you can see how it looks for:
 - [Failed build](https://github.com/dorny/FluentValidation/runs/1802163806?check_suite_focus=true)
 - [Successful build](https://github.com/dorny/FluentValidation/runs/1809435942?check_suite_focus=true)

Edit:
It's configured to run only when previous step (e.g. tests) already failed so it wouldn't affect successful builds at all.

Compared to scrolling through build step logs this action makes it somewhat easier to see which tests failed and why. It's useful mostly when test fails only in CI environment.

I created [test-reporter](https://github.com/dorny/test-reporter) mostly for my own needs but I would be glad if others found it useful too. During development I test it against few randomly picked popular OSS projects. FluentValidation was one of them so basically that's why I'm creating this PR now when the action hit 1.0 version.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,3 +40,12 @@ jobs:
       shell: pwsh
       env:
         DOTNET_SKIP_FIRST_TIME_EXPERIENCE: 1
+
+    - name: Create test report
+      if: success() || failure()
+      uses: dorny/test-reporter@v1
+      with:
+        name: Test results (${{ matrix.os }})
+        path: src/**/TestResults/*.trx
+        reporter: dotnet-trx
+        list-tests: failed
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,7 +42,7 @@ jobs:
         DOTNET_SKIP_FIRST_TIME_EXPERIENCE: 1
 
     - name: Create test report
-      if: success() || failure()
+      if: failure()
       uses: dorny/test-reporter@v1
       with:
         name: Test results (${{ matrix.os }})
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,11 +41,8 @@ jobs:
       env:
         DOTNET_SKIP_FIRST_TIME_EXPERIENCE: 1
 
-    - name: Create test report
-      if: failure()
-      uses: dorny/test-reporter@v1
+    - name: Upload test results
+      uses: actions/upload-artifact@v2
       with:
-        name: Test results (${{ matrix.os }})
+        name: test-results-${{ matrix.os }}
         path: src/**/TestResults/*.trx
-        reporter: dotnet-trx
-        list-tests: failed
--- a/.github/workflows/test-reporter.yml
+++ b/.github/workflows/test-reporter.yml
@@ -1,0 +1,19 @@
+name: Test Report
+
+on:
+  workflow_run:
+    workflows: ["CI"]
+    types:
+      - completed
+
+jobs:
+  report:
+    runs-on: 'ubuntu-20.04'
+    steps:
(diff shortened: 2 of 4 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/FluentValidation/FluentValidation?pullRequest=1633) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/FluentValidation/FluentValidation?pullRequest=1633) before we can accept your contribution.<br/><br/>:x: dorny [sign now](https://cla.dotnetfoundation.org/FluentValidation/FluentValidation?pullRequest=1633)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/FluentValidation/FluentValidation?pullRequest=1633) it.</sub>

dorny:
It failed when it tried to create the report using GitHub API. I assume it happened because the build was triggered from forked repo and token didn't have enough privileges. This was the first time I used it in PR to repo I didn't own. Well, at least I know about this issue now. I will look into it.

dorny:
For now I changed it so report is only created for failed builds. It still won't work for PRs from external collaborators but it doesn't matter much because report step would be executed only when build already failed. If workflow is triggered by maintainers of this repo and some test fails you will get the report.

JeremySkinner:
Hi, thanks for wanting to contribute this - it's a nice idea!

> It still won't work for PRs from external collaborators

I think this would be an issue for us. The report would be most useful for external contributors who are making PRs to the project to easily see which of their tests failed (3rd party contributors are less likely to run the tests themselves first, or to take the time to dig through the full build output in CI). If this isn't possible then I think it'd be of limited usefulness to the project. 

If you're able to resolve this, then I think it'd be great to have in the project, but if not then it's probably not something that will add much additional value.

dorny:
Yeah, makes sense. I have some idea I need to explore. I will post here updated in a couple of days.

dorny:
The trick is to upload test results (.trx files) as artifacts and process them in separate workflow triggered by [workflow_run](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_run) event.
Second workflow will be executed within context of default branch of this repository (instead of my fork) and will have write permission which we need for creating check run and annotations.

This is considered safe because only workflow defined by you has write access. Otherwise anybody could create PR and do some bad things with your access token. This pattern has to be used for any action which interacts with PR - adding comments or check runs with code annotations like my test-reporter.

On my side I have to update the action so it can find out URL to download and unzip artifacts and few other modifications.
I will create a test repo owned by different account to demo it.

dorny:
Now it also works for PRs from external collaborators.
`CI` workflow uploads test results (`.trx` files) as artifacts.
Afterwards separate workflow is triggered. It runs in the context of `main` branch with read/write access token.
Second workflow will download test results and create check run with test report in the original PR.
Explanation and example are here: [recommended-setup-for-public-repositories](https://github.com/dorny/test-reporter#recommended-setup-for-public-repositories)

Workflows triggered on `workflow_run` event must be defined on `main` branch so you won't actually see test report being created in this PR. It will start working for next PRs after this one is merged - if you are still interested  :)

To test it I created account `dorny-test` and copy of your repository. Afterwards I forked it using my regular account and created PR. By purpose I also make some test failing. You can see it here:
https://github.com/dorny-test/gh-api-test/pull/2/checks?check_run_id=1942153418#r3s21

PR #1636, opened 2021-02-07
Use IComparer for IncusiveBetween validator

virzak:
There are situations when property validation is made on a class which doesn't implement IComparable. The validation is usually a rare one and for that reason, IComparer is preferred to implementing IComparable.

This PR attempts to compare by street number - a custom comparer for odd situations.

If the concept is acceptable, will also implement for ExclusiveBetween.


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/FluentValidation/FluentValidation?pullRequest=1636) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/FluentValidation/FluentValidation?pullRequest=1636) before we can accept your contribution.<br/><br/>:x: virzak [sign now](https://cla.dotnetfoundation.org/FluentValidation/FluentValidation?pullRequest=1636)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/FluentValidation/FluentValidation?pullRequest=1636) it.</sub>

JeremySkinner:
Hi, thanks for the suggestion - I think it's a good idea idea. There's a few things things that will need to be changed:

Property validators are receiving some major changes for the 10.0 release (mainly that they're going to become generic), so this will need some work in order to make it compatible. If you want to have a go at rebasing this on top of the 10.0 branch then that would be great, as I'm not planning on adding new features to the 9.x series of releases. 

Secondly, the comparer property should not be optional on the new overload. If the property doesn't implement IComparable directly then the comparer should be required. 

Thirdly I think we should accept `IComparer<T>` rather than `IComparer` as the parameter type (as the InclusiveBetweenValidator is generic in 10.0 anyway, this should be easier to implement, although some work will need to be done with the generic constraints on the class as at the moment it's constrainted to `IComparable<TProperty>` in the 10.0 branch) 

Also you'll need to sign the CLA.

Thanks for wanting to be involved - let me know if you need me to clarify any of the above.

virzak:
It is clear. Didn't check out the 10.x branch, but it rebased nicely. Will finish up soon and let you know.

src/FluentValidation/Validators/RangeValidator.cs
```diff
@@ -0,0 +1,63 @@
...
+// limitations under the License.
+//
+// The latest version of this file can be found at https://github.com/FluentValidation/FluentValidation
+#endregion
+
+namespace FluentValidation.Validators {
+	using System;
+	using System.Collections.Generic;
+
+	public abstract class RangeValidator<T, TProperty> : PropertyValidator<T, TProperty>, IBetweenValidator {
```
virzak, line 23: @JeremySkinner The common code between Inclusive and Exclusive validator increased when dealing with IComparer<T>, so it made sense to factor it out a bit.

src/FluentValidation/Validators/RangeValidator.cs
```diff
@@ -0,0 +1,63 @@
...
+
+		object IBetweenValidator.From => From;
+		object IBetweenValidator.To => To;
+
+		protected int Compare(TProperty a, TProperty b) {
+
+			// Use explicitComparer first
+			if (explicitComparer != null)
+				return explicitComparer.Compare(a, b);
+			else if (a is IComparable comparableA)
```
JeremySkinner, line 57: Using the non-generic `IComparable` here is going to cause unnecessary boxing when working with value types - it would be better if we could find a way to use `IComparable<TProperty>` instead, as per the previous implementation.

JeremySkinner: It could be that we have to have 2 versions of the validators - one for handling `IComparable<T>` and one with the comparer.

JeremySkinner: The other option would be to always work with `IComparer<T>` and have a comparer implementation that explicitly wraps comparing two `IComparable<T>` instances.

JeremySkinner: Eg: 
```csharp
	public class ComparableComparer<T> : IComparer<T> where T : IComparable<T> {
		public int Compare(T x, T y) {
			return x.CompareTo(y);
		}
	}

```

This could then be passed in via the extension methods that explicitly work with IComparable.

src/FluentValidation/Validators/RangeValidator.cs
```diff
@@ -0,0 +1,63 @@
...
+				return comparableA.CompareTo(b);
+			else
+				throw new NotSupportedException("Object should either implement IComparable or IComparer should be passed");
```
JeremySkinner, line 60: If we go with my above suggestion then this code path can be removed

src/FluentValidation/Validators/RangeValidator.cs
```diff
@@ -0,0 +1,63 @@
...
+			From = from;
+
+			explicitComparer = comparer;
+
+			if (comparer.Compare(to, from) == -1) {
+				throw new ArgumentOutOfRangeException(nameof(to), "To should be larger than from.");
+			}
+		}
+
+		readonly IComparer<TProperty> explicitComparer = null;
```
JeremySkinner, line 44: Underscore prefix for private fields please, and could it be moved to the top of the class above the constructors. Thanks!

src/FluentValidation/Validators/RangeValidator.cs
JeremySkinner: Good idea, thanks! A couple of changes regarding handling of IComparable (see comments below)

src/FluentValidation/Validators/RangeValidator.cs
```diff
@@ -52,9 +52,9 @@ public abstract class RangeValidator<T, TProperty> : PropertyValidator<T, TPrope
 		protected int Compare(TProperty a, TProperty b) {
 
 			// Use explicitComparer first
-			if (explicitComparer != null)
-				return explicitComparer.Compare(a, b);
-			else if (a is IComparable comparableA)
+			if (_explicitComparer != null)
+				return _explicitComparer.Compare(a, b);
+			else if (a is IComparable<TProperty> comparableA)
```
JeremySkinner, line 57: This won't avoid boxing if `a` is a value type. This might seem pedantic but it does make a difference to performance under large workloads, and as part of the 10.0 release I'm trying to remove as much boxing as possible. 
I think having a comparer implementation as in my other comment would be better. 

If you don't have the availability to do this then let me know and I'll pick it up when I get some time.

src/FluentValidation/Validators/RangeValidator.cs
virzak: Went through a few iterations and I think this solution is not too bad. There are no breaking changes.

Both `ComparableComparer` and 2 versions had to be implemented to accommodate direct instantiation ```new InclusiveBetweenValidator<Person, int>(1, 10);``` calls.

Because of 4 different validators, I moved `IsValid` implementation down to `RangeValidator` and introduced HasError for the check. Not so sure about it as it could be confusing.

Some other thoughts/considerations:
- In my project there is a need to store degrees. The valid range is [0, 360) - meaning 0 is inclusive, but 360 isn't. If that's a going to be a popular request in the future, that means 2 more validators.
- Perhaps RangeValidator could take an Enum Inclusive, Exclusive, InclusiveFromExclusiveTo (for lack of a better name)
- Should be a separate issue probably: The "Value" argument is different from "PropertyValue" in the [reference](https://docs.fluentvalidation.net/en/latest/built-in-validators.html#inclusivebetween-validator)

JeremySkinner: Thanks, I'll have another review of this tomorrow. I might end up making a few tweaks and I'll let you know if I do.

>  If that's a going to be a popular request in the future, that means 2 more validators.

To be honest I probably wouldn't bother adding new classes for this, as it's easy enough to use `Must` to perform the comparison in this case.  

> "Value" argument is different from "PropertyValue" in the reference

These are different - {PropertyValue} is automatically added to all validators. The {Value} placeholder is the comparison value (and should actually be called {ComparisonValue} for consistency with other validators). I'll open a separate issue to rename this and add it to the docs.

JeremySkinner: Actually I was wrong about ComparisonValue - the {Value} placeholder in these 2 validators is a duplication of {PropertyValue}, so it should be possible to remove it, but I may leave it in place for backwards compatibility

JeremySkinner:
I made a few minor changes:
- Introduced a singleton property for the comparer instance
- Add some missing doc comments
- Make HasError protected
- Also decided that I preferred not having the distinction between InclusiveBetween/InclusiveBetweenWithComparer so removed the intermediary classes. Created some factory methods for instantiating the IComparable versions (I don't mind the breaking change here as this is a major version release anyway, plus it's uncommon for users to instantiate property validator classes directly).

JeremySkinner:
Squashed and merged

Issue #1147, opened 2019-06-05
Default display name splitting should handle properties containing periods


JeremySkinner:
`RuleFor(x => x.Address.Line1)` generates a display name of "Address. Line1" instead of "Address Line1".

SeanKilleen:
I thought I'd pick this up as a way of knocking something out. 

Is this still the case? I tried to write some failing tests in `PropertyChainTests` and a generic validation test against `Address.Line1` that checked the PropertyName, but I can't seem to recreate this behavior.

JeremySkinner:
It’s the display name that’s affected (ie the split version that appears inside the {PropertyName} error message template), not the actual value of PropertyName itself. I don’t have the availability to put together a failing test case atm but essentially I believe it’s reproducible with something like `RuleFor(x => x.Address.Line1).NotNull().WithMessage("{PropertyName}")`


...and then check the full error message

JeremySkinner:
Also this is extremely low priority as typically you’d specify an overridden display name by calling WithName in this scenario anyway. 

If you want to dive in then the documentation updates would be higher priority as I need to get those done before I can release 9.0

Frueber:
@JeremySkinner , I put together [something quick](https://github.com/FluentValidation/FluentValidation/pull/1638) for this that I hope gets us closer to a fix for this issue.
Let me know your thoughts, and if I messed something up with creating a PR to this project as well ;).

JeremySkinner:
Merged for 10.0

PR #1639, opened 2021-02-10
Updating string extension method SplitPascalCase to not return a singular delimiting period if preceding an upper case letter.

Frueber:
Updating string extension method SplitPascalCase to not return a singular delimiting period if preceding an upper case letter.

These changes stem from [this issue](https://github.com/FluentValidation/FluentValidation/issues/1147).

Note:
Another quick solution could be to use `Replace(". ", " ")` after the return value is built in `SplitPascalCase`.


JeremySkinner:
Thanks! I'll try and get this in within the next couple of days.

PR #1650, opened 2021-02-24
Update AsyncValidatorBase.cs todo

benmccallum:
Came to check if this had been done while I was upgrading to v9 and it looks like it wasn't so just more of just an FYI.


JeremySkinner:
Hi, this was postponed to a later version. 

This class has been removed and replaced by `AsyncPropertyValidator<T,TProperty>` in the 10.x branch.

benmccallum:
Awesome, thanks for the comment. I did just see 10.x release notes today with `AsyncPropertyValidator<T, TProperty>` and realised this would be modified. Looking forward to it!

PR #1668, opened 2021-03-18
Add netstandard2.0 back in FluentValidation.DependencyInjectionExtensions

mvestergaard:
Not sure if it was intentional or not, but this project should also support netstandard2.0 as long as the main library does.

And about this thing as a whole. In my opinion dropping support for netstandard2.0 in general is a bit premature.


JeremySkinner:
Thanks, yes they should both consistently target netstandard2.0 for now. 

> In my opinion dropping support for netstandard2.0 in general is a bit premature.

We'll be dropping it when .NET Core 2.1 ends LTS. FluentValidation 9.x will still be available for those that need to run on older frameworks.

PR #1694, opened 2021-04-13
#1501: ValidationFailure should be aware of it's RuleSet

### LINKED ISSUE #1501: ValidationFailure should be aware of it's RuleSet
**Is your feature request related to a problem? Please describe.**
When a `ValidationFailure` is constructed, it should be aware of the RuleSet name that it came from.

**Describe the solution you'd like**
Please add the following to `ValidationFailure`

```csharp
public string RuleSet { get; }
```

**Describe alternatives you've considered**
I've attempted to foreach all of my rulesets, but unfortunately I don't even see the validation failure when I do this.

```csharp
var failures = new Dictionary<string, ValidationResult>();

foreach (var ruleSet in _ruleSets)
{
    var result = await _validator.ValidateAsync(self, opts => opts.IncludeRuleSets(ruleSet));
    failures[ruleSet] = result; // unfortunately this is always IsValid = true;
    isValid = isValid && result.Errors.All(x => x.Severity != Severity.Error);
}
```

**Additional context**
Effectively I need to emit out validation failures based on the grouping of validation.
In other words, a particular segment of my overall object can be valid or invalid based on a RuleSet.

GuzzoLM:
Close #1501

### DIFF
--- a/src/FluentValidation.Tests/RulesetTests.cs
+++ b/src/FluentValidation.Tests/RulesetTests.cs
@@ -30,6 +30,17 @@ namespace FluentValidation.Tests {
 
 
 		[Fact]
+		public void ValidationFailure_displays_failing_ruleset() {
+			var validator = new TestValidator();
+			var person = new Person();
+			const string rulesetName = "Names";
+
+			var result = validator.Validate(person, options => options.IncludeRuleSets(rulesetName));
+
+			AssertFailed(result, rulesetName);
+		}
+
+		[Fact]
 		public void Executes_rules_in_specified_ruleset() {
 			var validator = new TestValidator();
 			var result = validator.Validate(new ValidationContext<Person>(new Person(), new PropertyChain(), new RulesetValidatorSelector(new[] { "Names" })));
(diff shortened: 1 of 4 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/FluentValidation/FluentValidation?pullRequest=1694) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/FluentValidation/FluentValidation?pullRequest=1694) before we can accept your contribution.<br/><br/>:x: GuzzoLM [sign now](https://cla.dotnetfoundation.org/FluentValidation/FluentValidation?pullRequest=1694)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/FluentValidation/FluentValidation?pullRequest=1694) it.</sub>

JeremySkinner:
Hi @GuzzoLM, thanks for wanting to contribute to the project! At the moment this implementation won't work correctly, so there are a few things that will need to be addressed before it can go into the project:

- This won't work with `RuleForEach` (changes have been made to `PropertyRule`, but not `CollectionPropertyRule`) 
- The async implementation is missing (changes have only been made to `InvokePropertyValidator` but not `InvokePropertyValidatorAsync`)
- The async tests don't work correctly - they don't include any async rules (`MustAsync`/`CustomAsync`) so don't actually test the async code path (see comment above)
- I'm not sure this will work proprely with child validators (`SetValidator`). I think this will end up overwriting the RuleSetsFailed on the failures generated by the child validators. I think there will be several combinations of this that will need carefully checking
- The tests don't check that the ruleset names are associated with the correct failure instances, only that there's at least one failure instance with the expected ruleset, which is too broad.
- Will need tests that check the behaviour when a rule is part of multiple rulesets

Additionally I'd like to see if it's possible to do this without looping over the failures. This is potentially very expensive for a large result set, so is something I'd like to avoid if possible considering this is very much an edge-case feature that will not be used by the majority of users. So this may need some further design discussions if we decide we do actually want this feature included.

JeremySkinner:
I'm going to close this as there's been no further activity here. Please feel free to re-open this if you want to resolve the issues that I raised.

PR #1733, opened 2021-05-11
Log Severity on ValidationException

jchannon:
This will log the severity of the rule when an exception is thrown


jchannon:
This is .NET OSS. I expect you to drop everything and merge this highly
critical PR in right now 😂😂

JeremySkinner:
Thanks!

PR #1765, opened 2021-06-12
🧹Code Cleanup

MrVitaly:
This PR contains the following minor changes:

- Fields that are only assigned in the constructor have been marked as 'readonly'.
Readonly fields can only be assigned in a class constructor. If a class has a field that's not marked as 'readonly' but is only set in the constructor, it could cause confusion about the field's intended use. To avoid confusion, it would be better to mark such fields 'readonly' to make their intended use explicit, and to prevent future maintainers from inadvertently changing their use.

- Redundant empty statements, unused 'using' directives, name qualifiers and boolean comparisons have been removed.
Although all that things won't change anything to the code, removing them:
	- Will help readability and maintenance;
	- Will help reduce the number of items in the IDE auto-completion list when coding;
	- May avoid some name collisions;
	- May improve compilation time (e.g. because the compiler has fewer namespaces to look-up when it resolves types).

### DIFF
--- a/src/CommonAssemblyInfo.cs
+++ b/src/CommonAssemblyInfo.cs
@@ -1,4 +1,3 @@
-using System;
 using System.Reflection;
 
 [assembly : AssemblyVersion("9.0.0.0")]
--- a/src/FluentValidation.AspNetCore/Adapters/ClientValidatorBase.cs
+++ b/src/FluentValidation.AspNetCore/Adapters/ClientValidatorBase.cs
@@ -16,12 +16,9 @@
 // The latest version of this file can be found at https://github.com/FluentValidation/FluentValidation
 #endregion
 namespace FluentValidation.AspNetCore {
-	using System;
 	using System.Collections.Generic;
-	using System.ComponentModel;
 	using Internal;
 	using Validators;
-	using System.Linq;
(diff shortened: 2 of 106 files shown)


JeremySkinner:
Hello, please don’t submit “code cleanup” pull requests. These don’t add any value and are hard to review. 

Pull requests should be targeted addres a single, specific issue and should be agreed with a project maintainer before being submitted to ensure they fit with the project roadmap: https://github.com/FluentValidation/FluentValidation/blob/main/.github/CONTRIBUTING.md#contributing-code

MrVitaly:
Thank you very much for the explanation! 🤝 I'll take it into account for the future.
Nevertheless, I fully understand that it can be very tough to review such large PR's, and this is why I have committed all these changes separately (see related [commits](https://github.com/FluentValidation/FluentValidation/pull/1765/commits)). There is only one a big commit (https://github.com/FluentValidation/FluentValidation/pull/1765/commits/de2b990442e5d53d1d21ecc6ad3e8fa16c718729) and it contains changes related only to 'using' directives and it probably won't take long to go over all those changes.
If I've convinced you that to review these changes are not so burdensome and they can be useful for this awesome project, you can return to the PR. 🤠

PR #1807, opened 2021-08-12
Deprecate NotEmpty in favour of Required

JeremySkinner:
- Introduce `Required`
- Deprecate `NotEmpty`
- Deprecate `Empty`
- Deprecate `Null`


JeremySkinner:
Decided against this because of the knock on implications

PR #1820, opened 2021-09-02
New method for test assertions: Only() (#1819)

### LINKED ISSUE #1819: New methods for testing validators: ShouldPass/Only

We are using FluentValidation quite intensively, and write lots of tests for our validators. That's why we came up with shorter and (I want to believe) more expressive syntax for asserting the behavior of our validators.

If you agree with the approach in general, I will go ahead and create a pull request so that more people could benefit from that syntax.

The current approach for positive assertions which is used in FluentValidation itself and can be considered recommended, looks like this:

```
    var result = validator.Validate(validaInstance);
```

The issue is, when the validation unexpectedly fails, this test does not give any information on what errors were raised, and we need to do manual debugging to find the reason.

For testing negative scenario, if we want to assert that only a single error was raised, we can use this approach:
```
    var result = validator.Validate(invalidInstance);
    result.Errors.Single().ErrorMessage.ShouldEqual("I am a bad guy!");
```

But again, in case of an unexpected error, there will be just `Sequence contains more than one element`, and we need to debug to find a culprit.

What we did was to add a couple of extension methods that check if the validation succeeded, verify that error messages match the expectation (when provided), and in case of a failure, output the detailed information about which errors were expected, which unexpectedly occurred, and which did not occur though expected.

In code, it looks like this:

```
    validator.TestValidate(validInstance).ShouldPass();
    validator.TestValidate(invalidInstance).ShouldFail(); // Here we do not care about the exact messages
    validator.TestValidate(invalidInstance).ShouldFail("The nose is too long"); // If we expect a single message
    validator.TestValidate(invalidInstance).ShouldFail("The nose is too long", "The eyes are too square", ...); // If we expect more than one message
```

The latter, with multiple error messages, is not of wide use because we usually test one rule in a single test, but still this can be useful. 

What this syntax does not support is verification of property names, error severity, etc. As well as it does not let you specify "I need to check that this error is raised, but there may be more, I don't care". But for this, the existing methods (`ShouldHaveValidationErrorFor` and similar) work perfectly, we are not proposing to do anything with them.

Initially, we used the approach very similar to what is used within FluentValidation for unit tests, but then switched to this new syntax as it is shorter and, well, more Fluent.

### Additional Context

_No response_

Aleksei-Pankratev-EPAM:
This PR adds an ability to assert that no other validation errors, except the ones that match specified conditions, were raised.

For this, we are adding a new method `.Only()` that can be used in various contexts:

```
    validator.TestValidate(x).ShouldHaveValidationErrorFor(x => x.Surname).Only();
    validator.TestValidate(x).ShouldHaveValidationErrorFor(x => x.Surname)
        .WithErrorMessage(message)
        .Only();
    validator.TestValidate(x).ShouldHaveValidationErrorFor(x => x.Surname)
        .WithErrorMessage(message)
        .WithSeverity(Severity.Warning)
        .Only();
    // and so forth
```

It cannot be used directly after `.Without*()` family of methods, because such record would be confusing and would have little sense:
```
    validator.TestValidate(x).ShouldHaveValidationErrorFor(x => x.Surname)
        .WithoutErrorMessage(message)
        .Only();
```

The changes required modifications of input and output types of `ValidationTestExtension` extension methods from `IEnumerable<ValidationFailure>` to `ITestValidationContinuation` or `ITestValidationWith`. These new interfaces do implement `IEnumerable<ValidationFailure>` so most of the existing code will not be affected but this is still a breaking change because some tests will be broken. For example, this does not work:
```
validator.TestValidate(new Person { }).Errors.WithoutErrorMessage(withoutErrMsg);
```

At the same time, it also has benefits of not polluting the shared namespace, because `IEnumerable<ValidationFailure>` is quite generic, and can be used in many places aside from unit tests.

In addition, I have refactored a few tests verifying failures by using try-catch block and asserting the exception was catched to use `Assert.Throws` instead.

----

Closes #1819

### DIFF
--- a/Changelog.txt
+++ b/Changelog.txt
@@ -6,6 +6,7 @@ Remove deprecated RunDefaultMvcValidationAfterFluentValidationExecutes option fr
 Remove deprecated Options property from RuleComponent.
 Remove deprecated TestHelper methods.
 Remove PropertyValidator backwards compatibility layer.
+Add method Only() asserting that no other validation errors were raised.
 
 10.3.1 - 
 Fix scale precision error message when digits is calculated as negative (#1790)
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -79,7 +79,21 @@ result.ShouldHaveValidationErrorFor(person => person.Name)
   .WithErrorCode("NotNullValidator");
 ```
 
-There are also inverse methods available (`WithoutMessage`, `WithoutErrorCode`, `WithoutSeverity`, `WithoutCustomState`)
+If you want to make sure no other validation failures occured, except specified by conditions, use method `Only` after the conditions:
+
+```csharp
+var result = validator.TestValidate(person);
+
+// Assert that failures only happened for Name property.
+result.ShouldHaveValidationErrorFor(person => person.Name).Only();
+
+// Assert that failures only happened for Name property and all have the specified message
+result.ShouldHaveValidationErrorFor(person => person.Name)
+  .WithErrorMessage("'Name' must not be empty.")
+  .Only();
+```
+
+There are also inverse methods available (`WithoutMessage`, `WithoutErrorCode`, `WithoutSeverity`, `WithoutCustomState`).
 
 ## Asynchronous TestValidate
 
--- a/src/FluentValidation.Tests/ValidatorTesterTester.cs
+++ b/src/FluentValidation.Tests/ValidatorTesterTester.cs
@@ -835,6 +835,138 @@ namespace FluentValidation.Tests {
 			});
 		}
 
+		[Fact]
+		public void ShouldHaveValidationErrorFor_Only() {
+			var validator = new InlineValidator<Person>();
+			validator.RuleFor(x => x.Surname).Must((x, ct) => false);
+			validator.TestValidate(new Person())
+				.ShouldHaveValidationErrorFor(x => x.Surname)
+				.Only();
+		}
+
+		[Fact]
+		public void ShouldHaveValidationErrorFor_Only_throws() {
+			var validator = new InlineValidator<Person>();
+			validator.RuleFor(x => x.Surname).Must((x, ct) => false);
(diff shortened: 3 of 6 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/FluentValidation/FluentValidation?pullRequest=1820) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/FluentValidation/FluentValidation?pullRequest=1820) before we can accept your contribution.<br/><br/>:x: Aleksei-Pankratev-EPAM [sign now](https://cla.dotnetfoundation.org/FluentValidation/FluentValidation?pullRequest=1820)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/FluentValidation/FluentValidation?pullRequest=1820) it.</sub>

JeremySkinner:
> So the first question is should I proceed with ITestValidationContinuation or you will never approve such change and it is not worth to spend time on it?

Yes that seems like a good approach to me - I was thinking something along the same lines. As it's a breaking change, please can you target the 11.x branch rather than main? 

> From what I heard yesterday, it seems, you probably will not support this initiative, giving preference to the syntax with braces. Am I right?

Yeah I'm not so keen on this one....let me have a think about it but my current inclination would be not to include this, partially because it increases the number of overloads but also because the assertions themselves are not async operations; only invoking the validator is async. But i'll think about it while I'm on holiday and let you know. For now I'd suggest keeping this PR focused on just adding `Only` and we can discuss this separately afterwards

Aleksei-Pankratev-EPAM:
@JeremySkinner this is ready for review at your convenience.

JeremySkinner:
Looks good, thanks. I've squashed this and merged in to 11.x

PR #1824, opened 2021-09-07
Fix MessageFormatterTets being dependent on current CultureInfo.

slightlysalty:
This is pretty self explanatory, tests in `MessageFormatterTets` are dependent on current `CultureInfo`. Might not be an issue on most build servers, but I find it disturbing that it can break anytime the machine configuration is changed.

### DIFF
--- a/src/FluentValidation.Tests/MessageFormatterTests.cs
+++ b/src/FluentValidation.Tests/MessageFormatterTests.cs
@@ -19,6 +19,7 @@
 namespace FluentValidation.Tests {
 	using System;
 	using System.Collections.Generic;
+	using System.Globalization;
 	using Internal;
 	using Xunit;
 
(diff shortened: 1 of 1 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/FluentValidation/FluentValidation?pullRequest=1824) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/FluentValidation/FluentValidation?pullRequest=1824) before we can accept your contribution.<br/><br/>:x: slightlysalty [sign now](https://cla.dotnetfoundation.org/FluentValidation/FluentValidation?pullRequest=1824)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/FluentValidation/FluentValidation?pullRequest=1824) it.</sub>

JeremySkinner:
Hi @slightlysalty thanks for this, but I believe this fixture would be unaffected by machine-specific configuration. It explicitly sets the culture in the constructor (the call to `CultureScope.SetDefaultCulture`), to force the tests to always use a specific culture, so they should never be affected by machine configuration. We do this in all of our tests that rely on any kind of culture-specific formatting, or specific languages.  Is this not working as you'd expect?

JeremySkinner:
I’m going to go ahead and close this as it breaks the tests, and also as per my previous comment it actually seems to work fine already regardless of machine culture

Issue #121, opened 2015-09-15
{PropertyName} adds spaces between words


jzamudio:
Hi 

I'm using version 5.6.2, I have a base class where I use custom error messages with {PropertyName} in them to have a uniform generic way to show the validation errors, my problem is that, for example, I have a property called "CountryCode", and in the message it shows the property name as "Country Code".

Is a way to configure the validator to respect the property names?, I know about the method .WithName but if I have to pass the name in all my rules the purpose of having a base class or use {PropertyName} is lost .

Thanks for your time

JeremySkinner:
So this is the default behaviour - when building error messages, PascalCased property names are split. You can override this by setting the global DisplayNameResolver property in your app startup routine: 

```
ValidatorOptions.DisplayNameResolver = (type, member, expr) => {
  return member.Name;
}
```

jzamudio:
Thanks for your help

PR #1828, opened 2021-09-13
Document ComparisonProperty placeholder

commonquail:
I discovered this placeholder accidentally, and after quite some time, while 1) trying to reference another property from a custom message, and 2) wanting to render the property names consistently. I want to render them consistently because to do otherwise invites unnecessary confusion.

---

I might take this opportunity to note that consistently referencing properties in custom messages is exceedingly painful:

- `ComparisonProperty` works perfectly for a select few validators but obviously does not work at all in any validator without first-class support for it; and so far it has been practically undocumented.
- I can't use `nameof(<property>)` because of the default display name resolver's insistence on transforming property names into not-property-names.
- I can't access the default display name resolver either, so I also can't `ddnr(nameof(<property>))`.
- Overriding the display name resolver with a bespoke version that could be used in this way is poorly documented, and specifically overriding it safely is not documented at all.

Concretely, this is completely impossible to do in `PredicateValidator`.

My wish would be for a built-in `PropertyPathDisplayNameResolver` that, for

```csharp
class AaAa { public string BbBb { get; set; } }
```

produces `AaAa.BbBb`, not `Aa Aa Bb Bb`.

Second to that, a documented, robust implementation of this in an accessible place so others can use it manually with less effort.

Third to that, a built-in `PropertyNameDisplayNameResolver` that robustly produces _just_ the property name, which I imagine is easier to program than the property path. Here is a decent version that recovers from null properties, which is a fault in the version posted in #121:

```csharp
ValidatorOptions.Global.DisplayNameResolver = (_, member, _) => member?.Name;
```

### DIFF
--- a/docs/built-in-validators.md
+++ b/docs/built-in-validators.md
@@ -44,6 +44,7 @@ Example error: *'Surname' should not be equal to 'Foo'*
 String format args:
 * `{PropertyName}` – Name of the property being validated
 * `{ComparisonValue}` – Value that the property should not equal
+* `{ComparisonProperty}` – Name of the property that the property should not equal, if any
 * `{PropertyValue}` – Current value of the property
 
 Optionally, a comparer can be provided to ensure a specific type of comparison is performed:
@@ -79,6 +80,7 @@ Example error: *'Surname' should be equal to 'Foo'*
 String format args:
 * `{PropertyName}` – Name of the property being validated
 * `{ComparisonValue}` – Value that the property should equal
+* `{ComparisonProperty}` – Name of the property that the property should equal, if any
 * `{PropertyValue}` – Current value of the property
 
 ```csharp
@@ -168,6 +170,7 @@ Notes: Only valid on types that implement `IComparable<T>`
 String format args:
 * `{PropertyName}` – Name of the property being validated
 * `{ComparisonValue}` – Value to which the property was compared
+* `{ComparisonProperty}` – Name of the property to which the property was compared, if any
 * `{PropertyValue}` – Current value of the property
 
 ## Less Than Or Equal Validator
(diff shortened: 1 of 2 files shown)


commonquail:
How do I sign the CLA as a specific email address, not as a general user?

JeremySkinner:
> How do I sign the CLA as a specific email address, not as a general user?

Hi, I'm not sure what you mean by this I'm afraid. The CLAs are handled by the .NET foundation so not something I can help with directly, but I believe it should let you sign in with your github account when you go to sign it (I believe it needs to be with the same github account you use to make the contribution)

> My wish would be for a built-in PropertyPathDisplayNameResolver...

Display name resolvers don't have access to the path information unfortunately. Building the path needs access to the current validation context in order to work out the hierarcy, but display name resolvers don't have access to this as they also need the option of being invoked in the clientside validation integration, where no context is available. 

I may be able to change this in a future version as there's only one place that actually relies on this, as far as I can see. 

For now you'd need to populate the path placeholder directly inside your predicate: 

```csharp
RuleFor(x => x.Foo).Must((model, propertyValue, context) => {
	context.MessageFormatter.AppendArgument("OtherPropertyPath", context.PropertyChain.BuildPropertyName("OtherProperty"));

	bool success = //...whatever
	return success;

}).WithMessage("referencing other property path: {OtherPropertyPath}");

```

(this could be cleaned up using an extension method).

commonquail:
> I believe it should let you sign in with your github account when you go to sign it (I believe it needs to be with the same github account you use to make the contribution)

That's unfortunate. I cannot and will not hand any and every future .NET Foundation project a blanket permission to sidestep the licensing under which I make a contribution. I submitted this contribution in accordance with the Apache 2.0 license; not, "... and whatever we decide on in the future, no questions asked".

> Display name resolvers don't have access to the path information unfortunately. Building the path needs access to the current validation context in order to work out the hierarcy

Oh, I thought they just extracted the information from the `Expression` with heuristics. Ah, well.

> For now you'd need to populate the path placeholder directly inside your predicate:

Thanks, this looks useful. I'll try that next time.

PR #1835, opened 2021-09-22
Update documentation of NotEmpty/Empty validators

athurn:
NotEmpty/Empty validators are working on IEnumerables:
https://github.com/FluentValidation/FluentValidation/blob/155cffc0d60113639865baafde757019b827415e/src/FluentValidation/Validators/NotEmptyValidator.cs#L36
https://github.com/FluentValidation/FluentValidation/blob/155cffc0d60113639865baafde757019b827415e/src/FluentValidation/Validators/EmptyValidator.cs#L36

This behaviour is currently not mentioned in the docs and caused a bit of confusion for us today :).

### DIFF
--- a/docs/built-in-validators.md
+++ b/docs/built-in-validators.md
@@ -16,7 +16,8 @@ String format args:
 * `{PropertyValue}` – Current value of the property
 
 ## NotEmpty Validator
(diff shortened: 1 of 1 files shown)


JeremySkinner:
Merged, thanks

PR #1874, opened 2021-12-17
Update AzerbaijaneseLanguage.cs to TwoLetterISOLanguageName

eminovperviz:
Hi guys .
I have localization problem.  I think that, finally  I found a solution. I m trying to get your localization logic. I notice that , In other langugages you have used TwoLetterISOLanguageName, but on Azerbaijani language is not the same like others.  So if ,I am not wrong, culture name should be change to "az".

ISO Codes List: https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes


JeremySkinner:
thanks

Issue #1704, opened 2021-04-17
Updated minimum supported platform to be .net6+


JeremySkinner:
We originally planned to drop netstandard2 in FluentValidation 10, but decided to hold off as .net platform convergence was delayed from .net 5 to .net 6. 

We should target .net 6 as our minimum supported platform. Support for netstandard2 and netstandard2.1 should be removed. 

Users who still require netstandard2 or netstandard2.1 support will continue to be able to use FluentValidation 11.x.

richardgavel:
What new APIs in netstandard 2.1 were you planning on consuming?

JeremySkinner:
Default interface methods is the main thing, which will help tidy up our fluent interface implementation. Additionally it sets a new baseline and ensures the new version can't be used on platforms that we don't support anymore (.net core 2.1 & net4x)

JeremySkinner:
Pushing back for another year

AgentFire:
[richardgavel](https://github.com/richardgavel), well, at least `Task` will be able to be switched for `ValueTask`.

PR #1894, opened 2022-02-09
Switch from Task to ValueTask

AgentFire:
Hi.

I've decided to suggest (via this merge request) switching all `Task` method results (and/or lambdas') into `ValueTask`. As it's said [here](https://blog.marcgravell.com/2019/08/prefer-valuetask-to-task-always-and.html), "always use ValueTask over Task", and however [here](https://devblogs.microsoft.com/dotnet/understanding-the-whys-whats-and-whens-of-valuetask/?ranMID=46131&ranEAID=a1LgFw09t88&ranSiteID=a1LgFw09t88-lqQBDxFLeGtsmfVLPqIecA&epi=a1LgFw09t88-lqQBDxFLeGtsmfVLPqIecA&irgwc=1&OCID=AID2200057_aff_7806_1243925&tduid=%28ir__1k329be3hokf6jneo33yzzo9bn2xt2gr9zx23jp900%29%287806%29%281243925%29%28a1LgFw09t88-lqQBDxFLeGtsmfVLPqIecA%29%28%29&irclickid=_1k329be3hokf6jneo33yzzo9bn2xt2gr9zx23jp900) Stephen Toub suggests only to use ValueTasks "when considered necessary".

I personally consider using ValueTasks in this library for the optimization reasons, but of course it is ultimately up to you to decide whether to accept this request.

Changes done:

1. Removed NETStandard2.0 support (as you have already planned to do, according to #1704).
2. Carefully switched from Task to ValueTask all around the sources. Each change was reviewed manually.

Comments are welcome.  
Thank you.

### DIFF
--- a/src/FluentValidation.DependencyInjectionExtensions/FluentValidation.DependencyInjectionExtensions.csproj
+++ b/src/FluentValidation.DependencyInjectionExtensions/FluentValidation.DependencyInjectionExtensions.csproj
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
     <PropertyGroup>
-        <TargetFrameworks>netstandard2.0;netstandard2.1</TargetFrameworks>
+        <TargetFrameworks>netstandard2.1</TargetFrameworks>
         <AssemblyName>FluentValidation.DependencyInjectionExtensions</AssemblyName>
         <PackageId>FluentValidation.DependencyInjectionExtensions</PackageId>
         <Product>FluentValidation.DependencyInjectionExtensions</Product>
--- a/src/FluentValidation.Tests.AspNetCore/ClientsideMessageTester.cs
+++ b/src/FluentValidation.Tests.AspNetCore/ClientsideMessageTester.cs
@@ -44,151 +44,151 @@ namespace FluentValidation.Tests.AspNetCore {
 		}
 
 		[Fact]
-		public async Task NotEmpty_uses_simplified_message_for_clientside_validation() {
+		public async ValueTask NotEmpty_uses_simplified_message_for_clientside_validation() {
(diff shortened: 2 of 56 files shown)


JeremySkinner:
Hi there,

Thanks for the suggestion. 

#1704 was moved to the backlog milestone so won't be part of the 11.0 release. 11.0 will continue to target netstandard2, so this won't be feasible for now. 12.0 would be the earliest we could consider implementing this, which would provisionally be targetting for sometime mid 2023, assuming that removal of netstandard2 is more palatable for consumers by then. 

I'm also not totally sure if introducing this breaking change is worth it. Have you run benchmarks to see what impact it has?

This is also targetting `main` (which is the 10.x release) rather than the `11.x` branch so would need rebasing, but it's probably not worth it at this point given the timelines I mentioned above. 

For future reference, please don't open PRs for large changes without first opening a feature request to discuss the changes with maintainers [as is mentioned in the contribution guidelines](https://github.com/FluentValidation/FluentValidation/blob/main/.github/CONTRIBUTING.md#contributing-code). Thanks!

cmeyertons:
@JeremySkinner  I'd like to second this change for using FluentValidation in high performance scenarios.  We have APIs capable of bulk uploading millions of rows in a streamed fashion and FluentValidation generates a lot of allocations.  

I can post my DotTrace findings once my firm renews my license!

JeremySkinner:
@cmeyertons this has already been implemented as of #1954 and is in the current releases. All of FluentValidation's internals now use `ValueTask` rather than `Task`. 

The top-level `ValidateAsync` method will remain as `Task`, but all of the internals have been switched over to ValueTask, which also allowed us to reduce a significant amount of duplication.

PR #1924, opened 2022-04-11
Count Grapheme Clusters (Emojis) once when calculating string length

### LINKED ISSUE #1923: Group Grapheme Clusters together when calculating the string length

[Grapheme clusters](https://www.unicode.org/glossary/#grapheme_cluster) (i.e. emojis) may be stored in a C# (unicode) string, but their lengths will be counted as two characters, the grapheme base and additional marks.

For intuitive use, it would be nice if FluentValidation took grapheme clusters into account, counting each exactly once.

Getting string length by value.Length (in LengthValidator.cs) could be changed to the value's StringInfo.LengthInTextElements:
``` csharp
var stringInfo = new System.Globalization.StringInfo(value);
int length = stringInfo.LengthInTextElements;
```

### Additional Context

N/A

CooperWolfe:
see Fixes #1923

### DIFF
--- a/src/FluentValidation.Tests/ExactLengthValidatorTester.cs
+++ b/src/FluentValidation.Tests/ExactLengthValidatorTester.cs
@@ -38,6 +38,14 @@ namespace FluentValidation.Tests {
 		}
 
 		[Fact]
+		public void When_the_text_contains_grapheme_clusters_then_the_validator_should_count_each_exactly_once()
+		{
+			var validator = new TestValidator { v => v.RuleFor(x => x.Surname).Length(4) };
+			var result = validator.Validate(new Person { Surname = "🌹🥀🌞🌚" });
+			result.IsValid.ShouldBeTrue();
+		}
+
+		[Fact]
 		public void When_the_text_length_is_smaller_the_validator_should_fail() {
 			var validator = new TestValidator {v => v.RuleFor(x => x.Surname).Length(10) };
 			var result = validator.Validate(new Person { Surname = "test" });
--- a/src/FluentValidation.Tests/ExactLengthValidatorTester.cs
+++ b/src/FluentValidation.Tests/ExactLengthValidatorTester.cs
@@ -38,8 +38,7 @@ namespace FluentValidation.Tests {
 		}
 
 		[Fact]
-		public void When_the_text_contains_grapheme_clusters_then_the_validator_should_count_each_exactly_once()
-		{
+		public void When_the_text_contains_grapheme_clusters_then_the_validator_should_count_each_exactly_once() {
 			var validator = new TestValidator { v => v.RuleFor(x => x.Surname).Length(4) };
 			var result = validator.Validate(new Person { Surname = "🌹🥀🌞🌚" });
 			result.IsValid.ShouldBeTrue();
--- a/src/FluentValidation.Tests/FluentValidation.Tests.csproj
+++ b/src/FluentValidation.Tests/FluentValidation.Tests.csproj
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp3.1;net5.0</TargetFrameworks>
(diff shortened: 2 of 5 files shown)


CooperWolfe:
Snuck into this PR is adding dotnet 6 to the target frameworks, which was necessary for me to run my tests on a machine without dotnet 5 runtime

src/FluentValidation.Tests/ExactLengthValidatorTester.cs
```diff
@@ -37,6 +37,14 @@ public class ExactLengthValidatorTester {
 			result.IsValid.ShouldBeTrue();
 		}
 
+		[Fact]
+		public void When_the_text_contains_grapheme_clusters_then_the_validator_should_count_each_exactly_once()
+		{
```
JeremySkinner, line 42: Please could move the opening brace be on the end of the preceding line ([as per our coding standard](https://github.com/FluentValidation/FluentValidation/blob/main/.github/CONTRIBUTING.md#contributing-code-and-content))? Thanks!

src/FluentValidation.Tests/FluentValidation.Tests.csproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp3.1;net5.0</TargetFrameworks>
+    <TargetFrameworks>netcoreapp3.1;net5.0;net6.0</TargetFrameworks>
```
JeremySkinner, line 3: Please can you remove this? This is already being handled as part of the 11.0 upgrade, and it's preferred that PRs focus on one specific change only.

src/FluentValidation.Tests/LengthValidatorTests.cs
```diff
@@ -105,6 +105,14 @@ public void When_the_text_is_exactly_the_size_of_the_lambda_lower_bound_then_the
 			result.IsValid.ShouldBeTrue();
 		}
 
+		[Fact]
+		public void When_the_text_has_grapheme_clusters_then_each_should_be_counted_exactly_once()
+		{
```
JeremySkinner, line 110: Brace layout again please

src/FluentValidation.Tests/StringEnumValidatorTests.cs
```diff
@@ -95,7 +95,7 @@ public class StringEnumValidatorTests {
 		public void When_enumType_is_not_an_enum_it_should_throw() {
 			var exception = Assert.Throws<ArgumentOutOfRangeException>(() => new TestValidator { v => v.RuleFor(x => x.GenderString).IsEnumName(typeof(Person)) });
 
-#if NETCOREAPP3_1 || NET5_0
+#if NETCOREAPP3_1 || NET5_0 || NET6_0
```
JeremySkinner, line 98: Please can this be reverted

JeremySkinner (changes requested):
Hi @CooperWolfe thanks so much for this! A couple of minor changes that i've commented on:

- Please ensure you match the project coding style for consistency (braces layout)
- Please can you revert the net6 targets that were added? It's best to keep PRs focused on a single issue. Feel free to add a separate PR for this, or otherwise I'll do it at some point as part of the 11.x branch.

Thanks!

JeremySkinner:
I'm also wondering whether this should be an opt-in behaviour, as technically this is a runtime breaking change. Let me have a think about this and will get back to you.

CooperWolfe:
Sure. Admittedly I had the same thought. Another validator entirely feels like overkill. A boolean flag feels shortsighted. A system-wide option feels inflexible. Maybe we introduce an overload using a configuration class? Something like:
```csharp
RuleFor(x => x.Content).Length(2, opt => opt.UseLengthInTextElements = true);
```

JeremySkinner:
I think a configuration object like that is probably overkill, I'd be tempted to go with an optional parameter for now (via a second overload of all of the `*Length` related methods). Although I am starting to wonder if it's easier to just point end users to using a custom rule with `Must` if they need to do this.

CooperWolfe:
That's fine, I suppose we can close this then

Issue #1942, opened 2022-05-24
Access to parent context


guffy1234:
### Is your feature request related to a problem? Please describe.

Hi
There is issue #904  that describes that the internal property PassThroughParentContext somehow (?) could be set to true to pass through parent context to children.
Why it still internal and how set it?
Why it is a bad idea to pass parent context all the time?

### Describe the solution you'd like

Remove this property and pass parent every time.

### Describe alternatives you've considered

Add yet another functor signature that will allow to give value for PassThroughParentContext

JeremySkinner:
The issue you linked to is from 2018 and no longer relevant - the property in question is no longer used.

guffy1234:
Hi @JeremySkinner 
I thinking about how to pass arbitrary data to validators via some context.
I found this example, 
https://gist.github.com/maca88/9e4551da22cf53056143bef76148a452
but quickly realized that it is based on pretty old FluentValidator library.
After that I found that there is dictionary RootContextData that mostly solves my needs. But the idea to use custom root validation context with typed properties have not leave me and I was trying to implement something like this with traversing by using ParentContext property. Then I studied your codes and found that it pretty difficult, almost impossible due to there is no legal way to enable ParentContext assignment. 
As for me I don't see any reason why it not just propagated by default. So I just tried to move your attention to this small issue. Mayby it's a time to change something.

JeremySkinner:
So in modern versions the `context` parameter is the same as `context.ParentContext` from older versions. 

For some background, older versions of FluentValidation (as per the example you linked to) received a `PropertyValidatorContext` which was a new/separate instance for each rule. ParentContext then allowed you to access the parent `ValidationContext<T>`. Modern versions of FluentValidation don’t work this way - PropertyValidatorContext no longer exists, you instead receive the `ValidationContext<T>` directly (which is a shared instance across all rules; this was done for performance reasons). 

Contexts are no longer created for each rule but are instead shared (for performance), and there is no longer a parent. New ones are now only created when using child validators (as the type parameters are different so the same instance can’t be reused). 

All that being said, RootContextData is the correct way of doing what you want.

guffy1234:
I was so quick so even created a pull request )
I think that the link between child context and parent from which it had been forked is useful. IMHO. Or all useless codes shall be removed just because they only confuse. 
But it's up to you of course. You may ignore the PR.

[...]

PR #1945, opened 2022-05-24
Restore linking between chaildren and parent contexts

guffy1234:
Hi

As I told you in #1942 it would be nice to just restore the link between parent context and children contexts. This will allow to creation of custom root contexts and traverse to root from any child validator. Otherwise, the existence of ParentContext property is just senseless.
As you said it's a volunteer project, so I trying to help you with MR. :)
I removed the obsolete boolean property and change the code to propagate _parent value by default.

### DIFF
--- a/src/FluentValidation/IValidationContext.cs
+++ b/src/FluentValidation/IValidationContext.cs
@@ -238,7 +238,6 @@ namespace FluentValidation {
 		/// Creates a new validation context for use with a child validator
 		/// </summary>
 		/// <param name="instanceToValidate"></param>
-		/// <param name="preserveParentContext"></param>
 		/// <param name="selector"></param>
 		/// <returns></returns>
 		public ValidationContext<TChild> CloneForChildValidator<TChild>(TChild instanceToValidate, IValidatorSelector selector = null) {
--- a/src/FluentValidation/IValidationContext.cs
+++ b/src/FluentValidation/IValidationContext.cs
@@ -238,13 +238,14 @@ namespace FluentValidation {
 		/// Creates a new validation context for use with a child validator
 		/// </summary>
 		/// <param name="instanceToValidate"></param>
+		/// <param name="preserveParentContext"></param>
 		/// <param name="selector"></param>
 		/// <returns></returns>
-		public ValidationContext<TChild> CloneForChildValidator<TChild>(TChild instanceToValidate, IValidatorSelector selector = null) {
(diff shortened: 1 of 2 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/FluentValidation/FluentValidation?pullRequest=1945) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/FluentValidation/FluentValidation?pullRequest=1945) before we can accept your contribution.<br/><br/>:x: guffy1234 [sign now](https://cla.dotnetfoundation.org/FluentValidation/FluentValidation?pullRequest=1945)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/FluentValidation/FluentValidation?pullRequest=1945) it.</sub>

src/FluentValidation/IValidationContext.cs
```diff
@@ -238,14 +238,13 @@ public ValidationContext(T instanceToValidate, PropertyChain propertyChain, IVal
 		/// Creates a new validation context for use with a child validator
 		/// </summary>
 		/// <param name="instanceToValidate"></param>
-		/// <param name="preserveParentContext"></param>
 		/// <param name="selector"></param>
 		/// <returns></returns>
-		public ValidationContext<TChild> CloneForChildValidator<TChild>(TChild instanceToValidate, bool preserveParentContext = false, IValidatorSelector selector = null) {
```
JeremySkinner, line 244: This is a breaking change to the public API. The caller should be updated to pass `true` rather than changing this method.

JeremySkinner (changes requested):
Thanks for this, I need to do some testing to make sure there are no unexpected side affects of doing this, but in principle I think this should be fine. A few changes that'll be needed though:

- Changing the method signature of `CloneForChildValidator` is a breaking change, which should be avoided outside of major version releases. This change should be reverted as it's breaking the build. 
- The calling method (`ChildValidatorAdaptor.CreateNewValidationContextForChildValidator`) should be updated to pass `true` as the second argument rather than removing the argument entirely.
- Please make sure you sign the CLA too or we won't be able to accept the PR 

Thanks for wanting to be involved!

JeremySkinner (approved):

JeremySkinner:
Merged, thanks. I'll aim to push out a release on Friday.

JeremySkinner:
Released in 11.0.2

PR #2003, opened 2022-09-07
add property resolver for include properties

sergeivalko:
Fix

This pull request fixes how the resolver works with `IncludeProperties`.
For example we have property resolver camelCase and when we are using 
`validator.Validate(new Person(), strategy => strategy.IncludeProperties(x=>x.Surname))`
then validator will not validate our surname property cause our resolver resolve it as "surname"

`IncludeProperties` = "Surname" but resolver (RuleFor(x=>x.Surname)) = "surname" and it can't find the right property for validate

### DIFF
--- a/src/FluentValidation/Internal/MemberNameValidatorSelector.cs
+++ b/src/FluentValidation/Internal/MemberNameValidatorSelector.cs
@@ -79,6 +79,7 @@ public class MemberNameValidatorSelector : IValidatorSelector {
 			throw new ArgumentException($"Expression '{expression}' does not specify a valid property or field.");
 		}
 
-		return chain.ToString();
+		var propertyName = ValidatorOptions.Global.PropertyNameResolver(typeof(T), expression.GetMember(), expression);
+		return propertyName;
 	}
 }
--- a/src/FluentValidation/Internal/MemberNameValidatorSelector.cs
+++ b/src/FluentValidation/Internal/MemberNameValidatorSelector.cs
@@ -73,13 +73,12 @@ public class MemberNameValidatorSelector : IValidatorSelector {
 	}
 
 	private static string MemberFromExpression<T>(Expression<Func<T, object>> expression) {
-		var chain = PropertyChain.FromExpression(expression);
+		var propertyName = ValidatorOptions.Global.PropertyNameResolver(typeof(T), expression.GetMember(), expression);
 
-		if (chain.Count == 0) {
+		if (string.IsNullOrEmpty(propertyName)) {
 			throw new ArgumentException($"Expression '{expression}' does not specify a valid property or field.");
 		}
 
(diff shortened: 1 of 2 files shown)


sergeivalko:
@JeremySkinner Have a look, please

JeremySkinner:
Hi @sergeivalko thanks for the suggestion. 

Does this cause issues with including full chains inside the expression? Eg if you do `IncludeProperties(x => x.Address.Line1)`, does this still work correctly? I think this might break that scenario (although I'm not sure if we have any tests for that at the moment).

JeremySkinner:
Ah, no, ignore my previous comment. I was misremembering how PropertyNameResolver worked. I think this should be fine - I'll try and do a bit more testing on it later this week and will then try and do a release, probably on Friday.

sergeivalko:
@JeremySkinner yes everything is working fine I've just checked so looking forward to feedback thanks

sergeivalko:
@JeremySkinner any updates?

JeremySkinner:
No, I haven't had time to look at this any further. I only work on FluentValidation in my spare time, and real-life takes priority. Please be patient and I'll try and get to it as quickly as I can.

sergeivalko:
thank you

src/FluentValidation/Internal/MemberNameValidatorSelector.cs
```diff
@@ -79,6 +79,7 @@ public class MemberNameValidatorSelector : IValidatorSelector {
 			throw new ArgumentException($"Expression '{expression}' does not specify a valid property or field.");
 		}
 
-		return chain.ToString();
+		var propertyName = ValidatorOptions.Global.PropertyNameResolver(typeof(T), expression.GetMember(), expression);
+		return propertyName;
```
JeremySkinner, line 83: I think this method can be simplified now to remove the creation of the chain entirely. 

```csharp
	private static string MemberFromExpression<T>(Expression<Func<T, object>> expression) {
		var propertyName = ValidatorOptions.Global.PropertyNameResolver(typeof(T), expression.GetMember(), expression);

		if (propertyName == null) {
			throw new ArgumentException($"Expression '{expression}' does not specify a valid property or field.");
		}

		return propertyName;
	}
```

JeremySkinner (changes requested):
Thanks for doing this.

The `MemberFromExpression` can be simplified as the chain variable is no longer needed. If you could make that change then I'll get this merged. Thanks!

src/FluentValidation/Internal/MemberNameValidatorSelector.cs
sergeivalko: done

JeremySkinner (approved):

JeremySkinner:
I've pushed out 11.2.2 with this change

PR #2019, opened 2022-10-20
Fix SourceLink MSBuild flags

buvinghausen:
@JeremySkinner the SourceLink integration for FluentValidation is broken you can verify that in [NuGet Package Explorer](https://nuget.info/packages/FluentValidation/11.2.2)

Assuming you do you CI builds in GitHub Actions which it looks like you do then just re-release as 11.2.3 and those flags should turn green.

### DIFF
--- a/src/FluentValidation.DependencyInjectionExtensions/FluentValidation.DependencyInjectionExtensions.csproj
+++ b/src/FluentValidation.DependencyInjectionExtensions/FluentValidation.DependencyInjectionExtensions.csproj
@@ -17,7 +17,6 @@
         <EnablePackageValidation>true</EnablePackageValidation>
         <PublishRepositoryUrl>true</PublishRepositoryUrl>
         <EmbedUntrackedSources>true</EmbedUntrackedSources>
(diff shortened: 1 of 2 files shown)


JeremySkinner:
@buvinghausen thanks! I'll look at getting this into the next version. However please could you remove the `<DebugType>embedded</DebugType>` lines from the PR? These are are already present in the csproj so don't need to be added again

JeremySkinner:
Thanks! 

I think even with these changes the package will be marked as `Non deterministic` as my publishing workflow involves compiling a release tag on my workstation rather than pushing directly from github actions to nuget (so I can manually verify the package). 

Technically, the build is deterministic as it involves checking out a specific tag, so I guess I could always set `<ContinuousIntegrationBuild>true</ContinuousIntegrationBuild>`?

buvinghausen:
@JeremySkinner I will just change the condition tag from GitHub Actions to build in release mode that should suffice

JeremySkinner:
Ah yes that's a good way to handle it. Thanks!

PR #2023, opened 2022-10-31
Make Dutch translations more consistent.

dnperfors:
We found some inconsistencies and annoyances in the Dutch translation, so here is a PR to fix them for everyone :)


JeremySkinner:
thanks!

PR #2033, opened 2022-12-08
fix: MinimumLengthValidator supports null validation

zhenlei520:
MinimumLengthValidator supports null validation. When the parameter to be validated is null, the input should be considered as a null value, and the final result will be obtained after comparing with the minimum length, see unit test for details

### DIFF
--- a/src/FluentValidation.Tests/LengthValidatorTests.cs
+++ b/src/FluentValidation.Tests/LengthValidatorTests.cs
@@ -143,6 +143,12 @@ public class LengthValidatorTests {
 		result.Errors.Single().ErrorMessage.ShouldEqual("The length of 'Surname' must be at least 4 characters. You entered 3 characters.");
 	}
 
+	[Fact]
+	public void When_the_minlength_validator_fails_the_content_is_null_and_error_message_should_be_set() {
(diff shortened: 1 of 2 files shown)


JeremySkinner:
Thanks for the suggestion, but this isn't the correct behaviour and isn't something we'll be including. 

Null input should cause the validator to be skipped - it shouldn't be treated as a 0-length string. If you want to check for null you should be including a `NotNull` check within your rule chain.

The `StringLength` data annotation attribute within .NET itself behaves in the same way.

zhenlei520:
I am very glad to receive your reply, although `DataAnnotations` does not handle null, additional verification is required, but in fact, it does not conform to our habits

JeremySkinner:
Include a `NotNull` check as part of your rule chain: `RuleFor(x => x.Value).NotNull().MinimumLength(4)`

zhenlei520:
> Include a `NotNull` check as part of your rule chain: `RuleFor(x => x.Value).NotNull().MinimumLength(4)`

Hmm, I can understand what you mean, but it’s just a little strange that `MinimumLength` doesn’t check null

PR #2035, opened 2022-12-19
Optimize regex validator

benmccallum:
Probably no need to instantiate the regex if the value being null never means it gets executed.

### DIFF
--- a/src/FluentValidation/Validators/RegularExpressionValidator.cs
+++ b/src/FluentValidation/Validators/RegularExpressionValidator.cs
@@ -58,7 +58,7 @@ public class RegularExpressionValidator<T> : PropertyValidator<T,string>, IRegul
 
 	public override bool IsValid(ValidationContext<T> context, string value) {
 		if (value == null) {
-			return true	
+			return true;
 		}
 		
 		var regex = _regexFunc(context.InstanceToValidate);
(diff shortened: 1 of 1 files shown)


benmccallum:
Was just digging around with .NET 7's `GeneratedRegex` attribute and it's possible with the `Func<T, Regex>` overload to call the generated regex property which is pretty cool, but thought I'd point out a potential optimization here.

src/FluentValidation/Validators/RegularExpressionValidator.cs
```diff
@@ -57,9 +57,13 @@ public class RegularExpressionValidator<T> : PropertyValidator<T,string>, IRegul
 	}
 
 	public override bool IsValid(ValidationContext<T> context, string value) {
+		if (value == null) {
+			return true
```
JeremySkinner, line 61: missing semicolon

JeremySkinner:
hi @benmccallum thanks for this, although it doesn't build at the moment. Could you add in this missing semicolon then I'll get this merged. Thanks!

benmccallum:
Ha, I swear 99% of the time I edit a file for something small like this online in GitHub I miss something and it doesn't build 😆 

🤞🏻 it's ok now. and no worries :)

JeremySkinner:
Thanks!

PR #2057, opened 2023-01-23
Support indexers in expression trees when generating PropertyChain for simple cases.

JeremySkinner:
- [ ] Need some tests around normalising property names for collection expressions in the test helper 

Investigate adding the additions from https://github.com/FluentValidation/FluentValidation/pull/2360 to this PR if we decide to proceed with this feature.

### DIFF
--- a/src/FluentValidation/Internal/PropertyChain.cs
+++ b/src/FluentValidation/Internal/PropertyChain.cs
@@ -70,10 +70,33 @@ public class PropertyChain {
 		});
 
 		var memberExp = getMemberExp(expression.Body);
+		string indexer = null;
 
 		while(memberExp != null) {
-			memberNames.Push(memberExp.Member.Name);
-			memberExp = getMemberExp(memberExp.Expression);
+			string propertyName = memberExp.Member.Name;
+
+			if (indexer != null) {
+				propertyName += $"[{indexer}]";
+				indexer = null;
+			}
+
+			memberNames.Push(propertyName);
+
+			// Handle indexers.
+			if (memberExp.Expression is MethodCallExpression mce
+			    && mce.Method is {IsSpecialName: true, Name: "get_Item"}
+			    && mce.Arguments.Count == 1
+			    && mce.Arguments[0] is ConstantExpression ce) {
+
+				memberExp = getMemberExp(mce.Object);
+				if (memberExp != null) {
+						indexer = ce.Value?.ToString();
+				}
+			}
+			else {
+				memberExp = getMemberExp(memberExp.Expression);
+			}
+
 		}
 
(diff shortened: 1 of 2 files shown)


mathiasi:
Any updates on this? Ran into this issue today that "x => x.MyList[0]" was not supported

JeremySkinner:
@mathiasi currently no plans to include this in a release given that it's easy to workaround with the overloads that take a string. I may revisit it in the future but i'm not keen on adding complexity to the PropertyChain logic. My current thinking is it's good enough for the majority of cases, and the string-based overloads are acceptable for this use case.

src/FluentValidation/Internal/PropertyChain.cs
```diff
@@ -70,10 +70,31 @@ public static PropertyChain FromExpression(LambdaExpression expression) {
...
 		var memberExp = getMemberExp(expression.Body);
+		string indexer = null;
 
 		while(memberExp != null) {
-			memberNames.Push(memberExp.Member.Name);
-			memberExp = getMemberExp(memberExp.Expression);
+			string propertyName = memberExp.Member.Name;
+
+			if (indexer != null) {
+				propertyName += "[" + indexer + "]";
```
muminvarici, line 79: String interpolation is recommended for next developments

src/FluentValidation/Internal/PropertyChain.cs
```diff
@@ -70,10 +70,31 @@ public static PropertyChain FromExpression(LambdaExpression expression) {
...
+				indexer = null;
+			}
+
+			memberNames.Push(propertyName);
+
+			// Handle indexers.
+			if (memberExp.Expression is MethodCallExpression mce) {
+				if (mce.Method is {IsSpecialName: true, Name: "get_Item"} && mce.Arguments.Count == 1 && mce.Arguments[0] is ConstantExpression ce) {
```
muminvarici, line 87: you can combine this and the previous if to decrease the code indentation level

JeremySkinner:
From #2360: 

> Bug 2 — Asymmetric normalization in TestValidationResult
> 
> ShouldHaveValidationError strips indexers from the actual error PropertyName before comparing, but not from the expected propertyName resolved from the expression. After fixing Bug 1, the resolved name contains [0] while the normalized actual name does not — so they still fail to match.
> 
> Bug 3 — ArgumentNullException on model-level rules (uncovered by fixing Bug 2)
> 
> Passing both sides through NormalizePropertyName exposed a crash when propertyName is null (model-level rules where PropertyName resolves to null).
> 

Investigate adding the additions from #2360 to this PR if we decide to proceed with this feature.

PR #2076, opened 2023-02-22
Change lambda parameter name to order

arphox:
A small improvement to an example code – the lambda parameter is named `orders` despite that it's a single order. I renamed it to `order` to emphasize that it is indeed **one** order.


JeremySkinner:
thanks

PR #2077, opened 2023-02-22
Reword version 9 warning about default comparison change

arphox:
I reworded the warnings about the version 9 default comparison mode, because version 9 is not in the future anymore, so referring to it with "will" may be misleading.


JeremySkinner:
thanks

PR #2078, opened 2023-02-22
Add missing dots and new lines

arphox:
And also add the "Example" words where it was omitted for the sake of consistency.


JeremySkinner:
thanks

PR #2088, opened 2023-03-09
Remove incorrect accents.

pipoburgos:
In Spanish, the plural of 'carácter' is 'caracteres' without an accent.


JeremySkinner:
Thanks for this. Could you fix up the indentation? It looks like all the tabs have been replaced with spaces, it'd be great if that could be reverted. Thanks!

PR #2111, opened 2023-06-01
Update async.md

knightmeister:
Added a note to call out that `ValidateAsync` will run both synchronous and asynchronous rules.


knightmeister:
@dotnet-policy-service agree

PR #2146, opened 2023-08-27
Added additional Dependency Injection (ServiceCollection) functions

mark-pro:
- Created additional connivance functions to register `AbstractValidators` to `IServiceCollections`.
- Updated documentation to include an example of registering a scoped validator using `AddValidator<T>`

@microsoft-github-policy-service agree

### DIFF
--- a/docs/aspnet.md
+++ b/docs/aspnet.md
@@ -1,17 +1,17 @@
 # ASP.NET Core
 
-FluentValidation can be used within ASP.NET Core web applications to validate incoming models. There are two main approaches for doing this: 
+FluentValidation can be used within ASP.NET Core web applications to validate incoming models. There are two main approaches for doing this:
 
 - Manual validation
 - Automatic validation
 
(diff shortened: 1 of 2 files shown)


mark-pro:
@microsoft-github-policy-service agree

JeremySkinner:
Hi, thanks for the suggestion and for wanting to be involved in the project, but I don't feel that these methods are necessary. The DI extensions are specifically intended to cover the case of locating all validators in a specified assembly, which isn't easy to do OOTB. Registering an individual validator can already be done easily with a direct call to `services.AddScoped` etc. 

For future reference please also take a look at our [contributor guidelines](https://github.com/FluentValidation/FluentValidation/blob/32befcfc9b87da25fe33f28c49aa6b6aab2b4ce6/.github/CONTRIBUTING.md#contributing-code) - we do ask that you open an issue to discuss any feature requests before opening a PR.

PR #2149, opened 2023-09-01
update tasks and Reference assemblies

SimonCropp:
i hit a minor dependency conflict pulling in different versions of these libs. 

not a bug with FV

i managed to workaround it, but figured it would be safe to update these libs anyway


JeremySkinner:
Merged, thanks. 12.x will be removing support for these frameworks anyway so these references will both be removed at that point.

SimonCropp:
@JeremySkinner happy with that

PR #2151, opened 2023-09-03
Chinese  language tags change

happtim:
The Language tags for Simplified Chinese and Traditional Chinese have been modified. The parent tag for zh-CN is zh-Hans, and the parent tag for zh-TW is zh-Hant.


JeremySkinner:
Thanks!

Issue #2152, opened 2023-09-04
NullReferenceException when property is null


MeikelLP:
### FluentValidation version

latest

### Summary

When you validate an object that may have nullable properties:

```cs
public class NullType {
	public List<int> List { get; set; }
}
```

and you validate their property without checking for null first

```cs
public class NullReferenceValidator : AbstractValidator<NullType> {
	public NullReferenceValidator() {
		RuleFor(x => x.List.Count).NotEmpty();
	}
}
```

It throws a generic `NullReferenceException` because the rule (`Expression` / `Func`) cannot be compiled. From what I read so far this is by design. However this is very hard to debug and to find out what property exactly is failing. I forked this repo and modified the `Create` in `PropertyRule` to include the expression in the closure. When you run you app in debug mode this massively helps you find out what property is null.

However this obviously has some downsides to:
* Performance due to increased allocations (I don't really know how to fix this here)
* Readability (may be improved somehow)

See repo for details. I even added a test to showcase this behaviour.

Is this the only way to debug this behaviour or is there another way?

### Steps to Reproduce

See repo:
https://github.com/FluentValidation/FluentValidation/compare/main...MeikelLP:FluentValidation:fix-null-debuggability

JeremySkinner:
Hi

> It throws a generic NullReferenceException because the rule (Expression / Func) cannot be compiled.

This isn't the case - the expression compiles successfully. However when it's invoked you'll get a NullReferenceException if the property is null, in the same way if you were to access it directly outside of a rule definition. When you handling nested properties you should always include an appropriate null check, in the same way you would with regular code too. 

>  I forked this repo and modified the Create in PropertyRule to include the expression in the closure

Could you provide an example of the output so I can see what difference this makes when debugging? Thanks

MeikelLP:
> This isn't the case - the expression compiles successfully. However when it's invoked you'll get a NullReferenceException if the property is null, in the same way if you were to access it directly outside of a rule definition. When you work with nested properties you should always include an appropriate null check, in the same way you would with regular code outside of fluentvalidation too.

Yes you are correct. The compilation does not fail but the invokation. Still this is totally fine for me but I just wanna know where it's failing not that it doesn't fail at all. I wanna know what rule/expression fails.

> Could you provide an example of the output so I can see what difference this makes when debugging? Thanks

Variables available in the scope where the exception is thrown (automatically provided by Rider when debugging/decompiling the lib):

Before:
[...]

PR #2153, opened 2023-09-08
Fix null debuggability

MeikelLP:
see #2152 

Please squash when merging :)


MeikelLP:
@dotnet-policy-service agree company="Konzept & Marketing GmbH"

JeremySkinner:
Thanks for this. A few comments:

- The try/catch logic should will need to be duplicated inside `CollectionPropertyRule` too
  - Will also need tests for handling collection rules (RuleForEach) 
- The exception type being thrown isn't correct - it should throw another NullReferenceException (not a ValidationException, which is only used for validation failures) which wraps the original, something like this:

```csharp
catch (NullReferenceException nre) {
  throw new NullReferenceException($"NullReferenceException occurred when executing rule for {context.PropertyPath}. If this property can be null you should add a null check using a When condition", nre);
}

```

MeikelLP:
I never really worked with `RuleForEach` but I tried my best :)

JeremySkinner:
That's great, thanks. Could you update the message in the exception thrown to be the one I suggested too? Once that's done I'll do a bit of extra testing and then should be able to get it merged.

MeikelLP:
done

JeremySkinner:
The logic and tests for RuleForEach wasn't correct - the catch was in the wrong place, it was catching collections thrown inside the property validator, not exceptions thrown by the rule definition. I've corrected this, and also done some additional cleanup. Merged for the 11.9 release.

PR #2155, opened 2023-09-12
Added Khmer langauge resouce.

sokheangkhun:
I have been added a new language which is Khmer language. So that I and others can use it in the projects.


sokheangkhun:
@dotnet-policy-service agree

JeremySkinner:
thanks!

PR #2174, opened 2023-11-22
fix memory leak enumerator

kronic:


JeremySkinner:
Hi, thanks for the suggestion.

This has the unintended sideffect of causing any uses of `System.Linq.IEnumerable<T>.Any` within the codebase to be switched to using this method in files where the `FluentValidation.Internal` namespace is imported (as this namespace is more specific). 

I've moved this to be a private non-extension method within the specific validators which resolves this and is a pattern I prefer as I don't think we need a globally available. method for this.

PR #2176, opened 2023-11-24
Upgrade to .net 8

JeremySkinner:


DanteDeRuwe:
@JeremySkinner any idea when this change will be part of a release? It is currently holding back our upgrade to .NET 8. Thanks in advance for your reply!

DanteDeRuwe:
@JeremySkinner Thanks for the quick response. I fixed the issue I had without knowing how and without any way to reproduce it. One of those days am I right? Have a nice one.

PR #2184, opened 2023-12-18
Register validators once per type

### LINKED ISSUE #2182: AddValidatorsFromAssemblyContaining would register Validators twice
### FluentValidation version

11.8.1

### ASP.NET version

(ASP).NET Core 8

### Summary

I honestly don't know if it is an issue, but it manifested as one to me when it occured.

2 (but actually any amount of) calls to `AddValidatorsFromAssemblyContaining(<assembly>)` would register every validator in the matching assembly 2 times. Clients, that retrieve something like `IEnumerable<IValidator<TTarget>>` via DI would get 2 instances of the same validator, potentially leading to executing it twice. Or thrice, or ... you get the point.

What sounds as an exotic use case in the first place, isn't too exotic for applications that make use of Vertical Slice Architecture. There the following structures are quite common:
- Feature A
-- Validators of Feature A (registered via FeatureA.ServiceCollectionExtensions.AddFeatureAServices)
- Feature B
-- Validators of Feature B (registered via FeatureA.ServiceCollectionExtensions.AddFeatureBServices)
- Feature C
-- ...

It basically boils down to the registrations in
https://github.com/FluentValidation/FluentValidation/blob/51e365b6cbe5a62f98d01f3b40e4fb06e6b1f65a/src/FluentValidation.DependencyInjectionExtensions/ServiceCollectionExtensions.cs#L95-L106
Here the descriptors are added via the `Add` method.
Changing those to `TryAdd`, that verifies if a corresponding service descriptor is already contained in the service collection, would resolve this issue.

But as said, I can't tell if this is an inteded behavior for a use case that I currently cannot see. In that case a parameter for example could provice both options:
``` cs
AddValidatorsFromAssemblyContaining(..., allowDuplicates: false) {
    if (allowDuplicates)
        services.Add(...)
    else
        services.TryAdd(...)
}
```

### Steps to Reproduce

```cs
public class ArbitraryTarget{}

public class ArbitraryValidator : AbstractValidator<ArbitraryTarget> { ... }

//Program.cs
services.AddValidatorsFromAssemblyContaining<Program>();
services.AddValidatorsFromAssemblyContaining<Program>();

//Somewhere.cs
var provider = services.BuildServiceProvider();
var validators = provider.GetServices<IValidator<ArbitraryTarget>>();
//Inspect validators.Count();
```

peterwurzinger:
Fixes #2182 

A few thoughts on that:
- I wrote unit tests, but had to reference `FluentValidation.DependencyInjectionExtensions.csproj` from the test project to not create a whole new project for one test class. But feel free to move it around as you like.
- Creating `ServiceCollection`s to test the behavior required a new package reference on `Microsoft.Extensions.DependencyInjection`. I referenced the newest version of that package (8.0.0), as it is compatible with netstandard2.0. This does not reflect the runtime environments of projects targeting .NET 6 or .NET 7, as corresponding versions of that package would be used by an enclosing framework version. But since it likely doesn't make a difference and to maintain readability, I didn't include the conditional TFM-madness in the project file. If that is desired, just ping me and I will submit it later.
- I tried to adhere to the coding style, naming and formatting that I saw. But also feel free to reformat as you like.

### DIFF
--- a/src/FluentValidation.DependencyInjectionExtensions/ServiceCollectionExtensions.cs
+++ b/src/FluentValidation.DependencyInjectionExtensions/ServiceCollectionExtensions.cs
@@ -19,6 +19,7 @@
 namespace FluentValidation;
 
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection.Extensions;
 using System;
 using System.Collections.Generic;
 using System.Reflection;
@@ -92,14 +93,14 @@ public static class ServiceCollectionExtensions	{
 		bool shouldRegister = filter?.Invoke(scanResult) ?? true;
 		if (shouldRegister) {
 			//Register as interface
-			services.Add(
+			services.TryAddEnumerable(
 				new ServiceDescriptor(
 					serviceType: scanResult.InterfaceType,
 					implementationType: scanResult.ValidatorType,
 					lifetime: lifetime));
 
 			//Register as self
-			services.Add(
+			services.TryAdd(
 				new ServiceDescriptor(
 					serviceType: scanResult.ValidatorType,
 					implementationType: scanResult.ValidatorType,
--- a/src/FluentValidation.Tests/DependencyInjectionExtensions/ServiceCollectionExtensionsTests.cs
+++ b/src/FluentValidation.Tests/DependencyInjectionExtensions/ServiceCollectionExtensionsTests.cs
@@ -1,0 +1,80 @@
+﻿#region License
+// Copyright (c) .NET Foundation and contributors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// The latest version of this file can be found at https://github.com/FluentValidation/FluentValidation
+#endregion
+
+namespace FluentValidation.Tests.DependencyInjectionExtensions;
+
+using Microsoft.Extensions.DependencyInjection;
+using System.Linq;
+using Xunit;
+
+public class ServiceCollectionExtensionsTests {
+
+	[Fact]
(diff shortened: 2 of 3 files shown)


src/FluentValidation.DependencyInjectionExtensions/ServiceCollectionExtensions.cs
```diff
@@ -92,14 +93,14 @@ public static IServiceCollection AddValidatorsFromAssemblyContaining<T>(this ISe
 		bool shouldRegister = filter?.Invoke(scanResult) ?? true;
 		if (shouldRegister) {
 			//Register as interface
-			services.Add(
+			services.TryAddEnumerable(
```
Tihomirov-Vasiliy, line 96: Why did you use TryAddEnumerable and not TryAdd here?

peterwurzinger: The desired behavior is to register every validator both
- as itself (serviceType == implementationType) to be able to resolve as itself
- as `IValidator<Target>` (serviceType = `IValidator<Target>`, implementationType = itself) to be able to inject `IEnumerable<IValidator<Target>>`

The first use case is covered via `TryAdd`, since it ensures that only one instance per serviceType is being registered. Multiple calls to AddValidatorsFromAssembly will therefore not lead to registering validators twice.
The second use case needs `TryAddEnumerable`, since that method matches the combination of serviceType and implementationType. As the name suggests, you can register multiple implementation types (e.g. `SomePartialModelValidator`, `SomeOtherPartialModelValidator`) as the same serviceType (e.g. `IValidator<Model>`).
`TryAdd(interfaceType, validatorType)` would lead to only the first validator for a validation target being registered and every other being ignored.

JeremySkinner: I think we need to be careful about using `TryAddEnumerable` here, typically you only want 1 validator to be registered for a specific type, so the behaviour of only having the first validator being registered and ignoring the rest might be preferable. What happens if you register multiple validators for the same type and then try and resolve just 1, does it return the first or throw an exception? If the former, then this is fine but if it throws then we'll need to use TryAdd instead

JeremySkinner: ok perfect, that should be fine then

lokrain:
Can I have permissions

JeremySkinner:
Merged, thanks

LeftTwixWand:
Spend 3 hours to understand, that IList is not supported... #2182

PR #2188, opened 2024-01-14
 Add overloading for WithErrorCode() method to be compatible with lambda function

meckbaig:
I added an overload to the WithErrorCode() method, similar to the WithMessage() method, which accepts a lambda function Func<T, string>.
I needed that to use custom runtime error codes through lambda functions.


meckbaig:
@dotnet-policy-service agree

JeremySkinner:
Thanks for the suggestion but this isn't something we'll be including. Error codes are intentionally not dynamic, and additionally this would be a breaking change. In future please discuss feature requests prior to opening a PR.

PR #2196, opened 2024-03-04
Added norwegian nynorsk language translation

ragnarstolsmark:
Added a language translation for Norwegian nynorsk. This will give nynorsk users a norwegian language translation since they would otherwise fallback to english.


JeremySkinner:
thanks!

PR #2199, opened 2024-03-12
Compile _expressionDebugView on #DEBUG only.

jscarle:
`_expressionDebugView` seems to be for debugging purposes only. `#DEBUG` has been added around it so caching can be reused between collection and non-collection accesses.


JeremySkinner:
Hi, this isn't correct. This isn't used in debugging, it's a necessary part of the key specifically to ensure that the accessor is *not* shared between collection and non-collection access. 

Collection and non-collection access *must* use different accessors or you'll get an exception at runtime.

If you run the tests with this change in place, you'll see the failures that this causes.

PR #2200, opened 2024-03-13
Improved AccessorCache<T>

jscarle:
As a follow up to #2199.

- Renamed internal field for clarity.
- Added additional comments.
- Eliminated closure during dictionary access for .NET5.0+.

### DIFF
--- a/src/FluentValidation.Tests.Benchmarks/AccessorCacheBenchmark.cs
+++ b/src/FluentValidation.Tests.Benchmarks/AccessorCacheBenchmark.cs
@@ -1,0 +1,58 @@
+#region License
+
+// Copyright (c) .NET Foundation and contributors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// The latest version of this file can be found at https://github.com/FluentValidation/FluentValidation
+
+#endregion
+
+namespace FluentValidation.Tests.Benchmarks {
+	using System;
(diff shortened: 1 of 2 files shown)


JeremySkinner:
Hi thanks for the suggestion. What kind of performance difference are you seeing with this change? I'm not seeing a difference with our regular benchmark suite.

jscarle:
I added an additional benchmark to specifically target the `GetCachedAccessor` method.

Here are the results of `GetOrAdd(key, _ => expression.Compile())`:
|                           Method |     Mean |    Error |   StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|--------------------------------- |---------:|---------:|---------:|-------:|------:|------:|----------:|
|                GetCachedAccessor | 80.90 ns | 0.334 ns | 0.312 ns | 0.0198 |     - |     - |     312 B |
| GetCachedAccessorWithCachePrefix | 86.95 ns | 0.720 ns | 0.674 ns | 0.0240 |     - |     - |     376 B |

And here are the results of `GetOrAdd(key, static (_, exp) => exp.Compile(), expression)`:
|                           Method |     Mean |    Error |   StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|--------------------------------- |---------:|---------:|---------:|-------:|------:|------:|----------:|
|                GetCachedAccessor | 73.19 ns | 0.757 ns | 0.708 ns | 0.0142 |     - |     - |     224 B |
| GetCachedAccessorWithCachePrefix | 80.53 ns | 0.311 ns | 0.276 ns | 0.0184 |     - |     - |     288 B |

In other words, 88 bytes less per method accessor. Not significant in itself, but over a large quantity of properties it would likely add up. There are probably more optimizations that could be done to reduce the memory usage of the cache, but I didn't dig that far into it.

JeremySkinner:
Thanks, I'm not really seeing a difference within the context of a full validation run, but every little helps. 

I'm not generally keen on conditional compilation unless there's a very good reason, so I think I'd prefer to do this as part of the next release (12.0) which has .net 6 as the minimum supported platform, meaning the conditional could be dropped. If you could rebase this on top of the 12.x-dev branch then that'd be great

jscarle:
The changes have been moved over to the 12.x-dev branch and the conditional compiler directive has been removed.

JeremySkinner:
Squashed & merged into 12.x, thanks!

Issue #2211, opened 2024-05-09
ScalePrecision validation error message is misleading


dropik:
### FluentValidation version

11.2.2

### ASP.NET version

.NET 6

### Summary

It appears that ```ScalePrecisionValidator``` produces an error message which does not quite reflect what is actually being validated.

I have a decimal property in my model class, which I want to be of precision ```10``` and scale ```2```. So in theory a value like ```123456789``` should be fine, since by not providing any zeros after decimal point the total precision of this number is ```9```. And it seems that ```ScalePrecisionValidator``` calculates precision correctly, but it still gives a validation error and a message like this:
```
'{PropertyName}' must not be more than 10 digits in total, with allowance for 2 decimals. 9 digits and 0 decimals were found.
```
Which sounds like a nonsense. After a bit of digging the FluentValidation source code it turned out what is actually being validated by ```ScalePrecisionValidator```:
```c#
public override bool IsValid(ValidationContext<T> context, decimal decimalValue) {
    var scale = GetScale(decimalValue);
    var precision = GetPrecision(decimalValue);
    var actualIntegerDigits = precision - scale;
    var expectedIntegerDigits = Precision - Scale;
    if (scale > Scale || actualIntegerDigits > expectedIntegerDigits) {
        context.MessageFormatter
	    .AppendArgument("ExpectedPrecision", Precision)
	    .AppendArgument("ExpectedScale", Scale)
	    .AppendArgument("Digits", actualIntegerDigits < 0 ? 0 : actualIntegerDigits)
	    .AppendArgument("ActualScale", scale);

	return false;
    }
    return true;
}
```
So it does not actually validate precision at all. What is being validated is that the integer part of the value must not overflow expected amount of digits, i.e. ```Precision - Scale``` amount of digits, which in my case was ```8``` digits. And now it makes sense why validation does not pass, because indeed my number has ```9``` digits in the integer part. But what is being communicated back to the user as validation error message and what is being validated are two different things. Without knowing this implementational detail this might be quite hard to the user to make out why validation does not pass.

This issue would actually relate to the abandoned issue #1200, but the proposal stated there would not resolve the problem I think, since in this case, for example, giving the total precision and giving amount of integer digits would result in the same number ```9```. So I would say that it might be better to communicate ```'ExpectedIntegerDigits'``` as something that was expected from a decimal value to have, and what is indeed validated. And the error message would sound something like:
```
[...]

PR #2213, opened 2024-05-14
Issue 2211. ScalePrecisionValidator error message clarification.

dropik:
As described in issue #2211, clarified `PrecisionScale` validator behaviour in docs. Provided a workaround to avoid self contradictory error messages. Also noted inconsistent russian translation of error message in confront to other translations.

### DIFF
--- a/src/FluentValidation/Validators/ScalePrecisionValidator.cs
+++ b/src/FluentValidation/Validators/ScalePrecisionValidator.cs
@@ -30,6 +30,9 @@ using System;
 /// Scale would be the number of digits to the right of the decimal point.
 /// Precision would be the number of digits. This number includes both the left and the right sides of the decimal point.
 ///
+/// It implies certain range of values that will be accepted by the validator.
+/// It permits up to Precision - Scale digits to the left of the decimal point and up to Scale digits to the right.
+///
 /// It can be configured to use the effective scale and precision
 /// (i.e. ignore trailing zeros) if required.
 ///
--- a/src/FluentValidation/Validators/ScalePrecisionValidator.cs
+++ b/src/FluentValidation/Validators/ScalePrecisionValidator.cs
@@ -57,11 +57,20 @@ public class ScalePrecisionValidator<T> : PropertyValidator<T, decimal> {
 		var actualIntegerDigits = precision - scale;
 		var expectedIntegerDigits = Precision - Scale;
 		if (scale > Scale || actualIntegerDigits > expectedIntegerDigits) {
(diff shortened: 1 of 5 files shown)


dropik:
@dotnet-policy-service agree
@dotnet-policy-service agree

JeremySkinner:
Thanks for putting together a PR @dropik. It'll take me some time to review this and consider the implications.

hi @nhowell I don't suppose you could take a look at this if you get some time? I know you did some work on the ScalePrecisionValidator before and would appreciate your input

src/FluentValidation/Validators/ScalePrecisionValidator.cs
```diff
@@ -57,11 +60,20 @@ public class ScalePrecisionValidator<T> : PropertyValidator<T, decimal> {
 		var actualIntegerDigits = precision - scale;
 		var expectedIntegerDigits = Precision - Scale;
 		if (scale > Scale || actualIntegerDigits > expectedIntegerDigits) {
+			// Precision and scale alone may be not enough to describe why actually a value is invalid,
```
nhowell, line 63:
```suggestion
			// Precision and scale alone may not be enough to describe why a value is invalid.
```

src/FluentValidation/Validators/ScalePrecisionValidator.cs
```diff
@@ -57,11 +60,20 @@ public class ScalePrecisionValidator<T> : PropertyValidator<T, decimal> {
 		var actualIntegerDigits = precision - scale;
 		var expectedIntegerDigits = Precision - Scale;
 		if (scale > Scale || actualIntegerDigits > expectedIntegerDigits) {
+			// Precision and scale alone may be not enough to describe why actually a value is invalid,
+			// for example for 123 and expected precision 3 and scale 2 the value is invalid, yet precision
```
nhowell, line 64:
```suggestion
			// For example, given an expected precision of 3 and scale of 2, the value "123" is invalid, even though precision
```

nhowell (changes requested):
I agree with the changes to the error message. Makes sense.

I provided some suggestions on the wording of a code comment that I found a bit clunky, but otherwise this looks good to me.

dropik:
@JeremySkinner , @nhowell , thank you for your time! @nhowell , yes this looks better, thank you for suggestion!

nhowell (approved):

JeremySkinner:
Thanks! I'll aim to get this merged in the next time I've got time to work on FV

PR #2229, opened 2024-07-06
StringEnumValidator, cache enum names

Xyzzyall:
This will prevent unnecessary assembly scans and allocations in `StringEnumValidator.IsValid`. 

Check the source code of `Enum.GetNames` to understand the reasons of this change.

### DIFF
--- a/src/FluentValidation/Validators/StringEnumValidator.cs
+++ b/src/FluentValidation/Validators/StringEnumValidator.cs
@@ -25,9 +25,9 @@ using FluentValidation.Internal;
 using Resources;
 
 public class StringEnumValidator<T> : PropertyValidator<T, string> {
-	private readonly Type _enumType;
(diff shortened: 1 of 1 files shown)


Xyzzyall:
@dotnet-policy-service agree

Xyzzyall:
Maybe there are reasons I don't know why `Enum.GetNames` should be used on every validation, which btw could be many on a high load.

If there is any, I wish to see some comments straight in source code on why this should be called on every validation.

JeremySkinner (changes requested):
hi @Xyzzyall this doesn't build. 

Line 39 is trying to reference a field that's been removed. It should be `_enumNames = Enum.GetNames(enumType);`

Please can you correct this, and then I'll merge it in.

JeremySkinner (approved):

JeremySkinner:
thanks!

PR #2238, opened 2024-08-31
Update DefaultValidatorExtensions.cs (Added NotNullAndEmpty)

ZaynTariq:
This commit introduces a new extension method NotNullAndEmpty for the IRuleBuilder in the validation framework. The method provides a convenient way to enforce both 'not null' and 'not empty' validation rules on properties.


JeremySkinner:
Thanks for the suggestion but this isn't something we'll be including. The syntax is already succinct enough that this doesn't justify the creation of a new method.

Issue #2240, opened 2024-09-02
Add async overload for Where on RuleForEach


PeterKaspar:
### Describe the solution you'd like

It would be nice if I could write something like this
```c#
RuleForEach(obj => obj.Items)
    .WhereAsync(async item => await CheckSomeAsyncConditionWhichNeedsExternalDependenciesAsync(item))
   .MustAsync(...);
```

PR #2241, opened 2024-09-02
Add WhereAsync as conterpart to Where on collection rules (#2240)

JeremySkinner:
Implements #2240

### DIFF
--- a/src/FluentValidation/DefaultValidatorOptions.cs
+++ b/src/FluentValidation/DefaultValidatorOptions.cs
@@ -398,6 +398,20 @@ public static class DefaultValidatorOptions {
 	}
 
 	/// <summary>
+	/// Applies an asynchronous filter to a collection property.
+	/// </summary>
+	/// <param name="rule">The current rule</param>
+	/// <param name="predicate">The condition</param>
+	/// <returns></returns>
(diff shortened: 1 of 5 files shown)


src/FluentValidation/DefaultValidatorOptions.cs
```diff
@@ -397,6 +397,20 @@ public static IRuleBuilderInitialCollection<T, TCollectionElement> Where<T, TCol
 		return rule;
 	}
 
+	/// <summary>
+	/// Applies an asynchronous filter to a collection property.
+	/// </summary>
+	/// <param name="rule">The current rule</param>
+	/// <param name="predicate">The condition</param>
+	/// <returns></returns>
+	public static IRuleBuilderInitialCollection<T, TCollectionElement> WhereAsync<T, TCollectionElement>(this IRuleBuilderInitialCollection<T, TCollectionElement> rule, Func<TCollectionElement, Task<bool>> predicate) {
```
kronic, line 406: can use ValueTask<bool>

JeremySkinner: The rest of the API uses `Task` and I prefer consistency

PR #2260, opened 2024-10-06
enable nullable in TrackingCollection

SimonCropp:


JeremySkinner:
Thanks for the suggestion but I'm not keen on the nullable feature so won't be including this

PR #2261, opened 2024-10-06
enable nullable in FluentValidation.DependencyInjectionExtensions

SimonCropp:


JeremySkinner:
Thanks for the suggestion but I'm not keen on the nullable feature so won't be including this

SimonCropp:
interesting. any reasoning so i can get more understanding of your perspective?

JeremySkinner:
Sure,

- I don't like the 'virality' of nullable, how it tends to cascade throughout a codebase as you start introducing it 
- I don't like how it introduces dual meaning of the `?` operator (ie how it means something different in the context of nullable value types vs nullable reference types) 

It's just a personal preference really

SimonCropp:
fair enough

PR #2264, opened 2024-10-06
Use central package management

hamidrezahy:


JeremySkinner:
I don't see a need for this, and it also conflicts with the package changes in the `12.x-dev` branch. In future please open an issue to discuss significant changes before making them, thanks

PR #2265, opened 2024-10-08
make RuleFor [Pure]

SimonCropp:
thoughts on making RuleFor `[Pure]`.

IDEs can then warn when the returned RuleBuilder is not used.

image


JeremySkinner:
Hmm I don't think this is a good use for Pure as `RuleFor` is not a pure function - it's creating rule state internally. Whilst I can appreciate the warning can be useful, it feels like it's abusing the attribute to do something it wasn't intended for, so I don't think I want to include this.

Issue #2276, opened 2024-11-08
Add WithState(...) overload which allows accessing ValidationContext<T>


ChThy:
### Is your feature request related to a problem? Please describe.

When using custom ValidationContexts, it is currently not possible to access addional properties on these ValidationContexts when using `WithState(...)` to customize the validation message.

### Describe the solution you'd like

Adding an overload of the `WithState(...)` method to `DefaultValidatorOptions` class which takes in a `Func<T, TProperty, ValidationContext<T>, object>` as stateProvider.

This seems like low hanging fruit since the ValidationContext is available there. Tried adding the function below to `DefaultValidatorOptions` class and everything builds and all tests run green (Code changes are ready and I am willing to open a pull request):
```cs
public static IRuleBuilderOptions<T, TProperty> WithState<T, TProperty>(this IRuleBuilderOptions<T, TProperty> rule, Func<T, TProperty, ValidationContext<T>, object> stateProvider) {
  stateProvider.Guard("A lambda expression must be passed to WithState", nameof(stateProvider));

  var wrapper = new Func<ValidationContext<T>, TProperty, object>((ctx, val) => {
    return stateProvider(ctx.InstanceToValidate, val, ctx);
  });

  Configurable(rule).Current.CustomStateProvider = wrapper;
  return rule;
}
```

### Describe alternatives you've considered

The alternative approach that I went for is a custom extension method which mimics `WithState(...)` and passes the ValidationContext<T> in the stateProvider argument. (Almost the exact same extension method as above)

JeremySkinner:
Thanks for the suggestion, that seems like a reasonable enhancement. If you could supply a PR along with appropriate tests then I'd be happy to review this and get it into a future release

ChThy:
@dotnet-policy-service agree

ChThy:
I seem to lack the permission to create a new branch. Am I overseeing anything?

JeremySkinner:
@ChThy github works by using _forks_. You need to fork the repository first, create the new branch in your fork, and then submit the pull request back to this repository from your fork. You don't create branches in this repo directly. 

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/fork-a-repo

ChThy:
@JeremySkinner Thanks! Didn't know that yet. 

Pull request is ready #2278

JeremySkinner:
Thanks! I'll aim to get this reviewed and merge the next time i'm working on FluentValidation

[...]

PR #2278, opened 2024-11-10
Add WithState(...) overload with access to ValidationContext<T>

ChThy:
#2276


ChThy:
@dotnet-policy-service agree

PR #2283, opened 2024-12-07
Add Serbian cyrillic language

rammba:
Hello @JeremySkinner, I'm adding Serbian cyrillic with these changes.

As you probably don't know, cyrillic version is kind of official language for us. All legal papers, documents, etc must be written on cyrillic. In that sense, I've added it here as well.

I must point potential **breaking change** here, because I moved usage of [sr](https://www.localeplanet.com/dotnet/sr/index.html) culture to newly created cyrillic language. As an additional explanation, all Serbians knows both cyrillic and latin perfectly. Basically it's the usage of the same letters, only on different letter groups.
Is this somehow problematic for your library?

I hope you will agree to add this change to the library. Please tell me if I missed to configure or change something.


src/FluentValidation/Resources/Languages/SerbianLatinLanguage.cs
```diff
@@ -22,29 +22,29 @@
 
 namespace FluentValidation.Resources;
 
-internal class SerbianLanguage {
-	public const string Culture = "sr";
+internal class SerbianLatinLanguage {
+	public const string Culture = "sr-Latn";
```
rammba, line 26: This can introduce breaking change, because `sr` culture will now be used on cyrillic version.

JeremySkinner:
hi @rammba thanks for this. I'm happy with this change, but breaking changes can only occur in major releases, so please could you re-target this PR to the `12.x-dev` branch? Thanks!

rammba:
Thanks for the input. I've changed base branch as you suggested. Do you maybe know when is planned to go live with v12?

JeremySkinner:
There's no planned release date at the moment as I currently have other priorities in life. If I get time over the christmas holidays I hope to pick it up again but I can't definitely promise that'll be the case.

rammba:
Thanks for the info. You can merge these changes then. I'm looking forward to see it live soon 🚀

rammba:
Hey @JeremySkinner, I just noticed that there is no migration to v12 document. Should it be added only for these, or also for some other changes on this branch?

PR #2289, opened 2024-12-23
Update including-rules.md

gopal-chinkere:
If the included validator is alphabetically/hierarchically placed below the root validator then only included validator is getting called and validated on the model while root is completely ignored. This issue was reported on the stackoverflow.com but nobody had given solution to this problem. I thought this small text edit might help developers to solve their problem with include() validator.

### DIFF
--- a/docs/including-rules.md
+++ b/docs/including-rules.md
@@ -41,5 +41,5 @@ public class PersonValidator : AbstractValidator<Person>
 
 ```eval_rst
 .. note::
-    You can only include validators that target the same type as the root validator.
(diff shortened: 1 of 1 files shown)


JeremySkinner:
This isn't correct.

Included validators are treated just the same as any other rules (included rules are conceptually similar to a call to RuleFor`(x => x).SetValidator(...)`). As with all rules, they are executed in the order they're defined.

If you believe that you're seeing a different behaviour please open a bug report with clear steps to reproduce.

PR #2294, opened 2025-01-04
Simplify guarding

Corniel:
By using `[CallerArgumentExpression]` (available for .NET standard via [PolySharp](https://github.com/Sergio0694/PolySharp)), and the introduction of `GuardNotNull` the plumbing needed to defend against null dereferences has hopefully improved.


JeremySkinner:
Thanks for the suggestion but these calls have already been replaced with `ArgumentNullException.ThrowIfNull` in the `12.x-dev` branch. This is preferable over using a 3rd party package. See #2009.

Corniel:
Note that PolySharp is a code generator package, not a runtime dependency.

PR #2295, opened 2025-01-04
Use decimal scale to simplify PrecisionScaleValidator

### LINKED ISSUE #2288: Suggestion: ScalePrecisionValidator might be simplified

Not a functional problem, but possibly a maintenance improvement - ScalePrecisionValidator has a lot of low-level code to check precision and scale. It is difficult to understand what is happening here:.

https://github.com/FluentValidation/FluentValidation/blob/c8c6400fa2f467d65669b78804613d44066c1c33/src/FluentValidation/Validators/ScalePrecisionValidator.cs#L99-L147

Part of the code might be replaced using the decimal.Scale property: https://stackoverflow.com/a/77213895/1114918

_No response_

### Additional Context

_No response_

Corniel:
After the discussion in #2288, I decided to try to simplify the implementation of `PrecisionScaleValidator<T>.IsValid()`. By using `decimal.Scale` and getting the scale and the precision in one go (instead of two) I was able to reduce the duration by 47%. I think the code is also easier to follow (which was the goal).

|  Method |       Mean |    Error |   StdDev | Ratio |
|-------- |-----------:|---------:|---------:|------:|
| 11.x | 1,654.6 us | 32.37 us | 28.69 us |  1.00 |
| 12.x |   875.9 us |  6.31 us |  5.27 us |  0.53 |

Note that these numbers are for 1000 validations.

Closes #2288


JeremySkinner:
thanks for this, I’ll aim to get it merged for 12.0-preview2

Corniel:
@JeremySkinner Do you want to keep the old implementation as a reference (as I currently did), or can it be removed from the benchmark project?

JeremySkinner:
It can be removed, thanks

Corniel:
Done.

JeremySkinner:
Thanks for doing this, I've merged it in.

For future reference, when contributing please keep pull requests focused on specific, individual change. In this case the reformatting of the tests makes the reviewer's job more difficult, and would've been preferable as a followup.

PR #2297, opened 2025-01-04
Align Func arguments of comparable rules

Corniel:
While working one some code, I noticed that some LT/GT rules had `Func<T, TProperty>` called `valueToCompare`, and other `expression`. I guess the latter was intended. This is a breaking change, but as you are working v12.x I guess it is perfectly fine to apply the change.

If you prefer to call arguments `valueToCompare` instead, that is fine with me. :)


JeremySkinner:
Thanks for this, I've merged it in.

For future reference please can you ensure that unrelated formatting changes aren't included in PRs as they make it harder to review the intended changes.

PR #2321, opened 2025-05-17
Enable IsAotCompatible.

henric-root:
Add attributes to items that should not be used in AOT builds.

Rework `EnumValidator` to support AOT builds without causing link errors along the lines of:
```
/_/src/FluentValidation/Validators/EnumValidator.cs(103): AOT analysis error IL3050: FluentValidation.Validators.EnumValidator`2.EvaluateFlagEnumValues(Int64,Type): Using member 'System.Enum.GetValues(Type)' which has 'RequiresDynamicCodeAttribute' can break functionality when AOT compiling. It might not be possible to create an array of the enum type at runtime. Use the GetValues<TEnum> overload or the GetValuesAsUnderlyingType method instead. [/src/xxxx/xxxx.csproj]
```

### DIFF
--- a/src/FluentValidation/AssemblyScanner.cs
+++ b/src/FluentValidation/AssemblyScanner.cs
@@ -23,12 +23,14 @@ namespace FluentValidation;
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Reflection;
 
 /// <summary>
 /// Class that can be used to find all the validators from a collection of types.
 /// </summary>
+[RequiresUnreferencedCode("Types might be removed")]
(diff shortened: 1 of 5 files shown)


JeremySkinner:
Thanks for the suggestion but this introduces several breaking changes so I'm afraid this isn't something that can be included outside of a major version release.

- Changes to the EnumValidator api surface are breaking
- Introduction of the NullableEnumValidator is incorrect; the wrong error message will be generated and there are no tests around its introduction.

In future please open an issue to discuss proposed changes before making them.

manuelelucchi:
@JeremySkinner I think this feature should be re-evaluated since
- Microsoft is pushing hard on it
- It's basically mandatory on some platform if you want decent performances (Maui, Wasm, aspnetcore using stuff like azure functions/aws lambdas)

PR #2334, opened 2025-07-03
Add Tamil (ta) Language

UTHAYANITHI:
## I have been added a new language which is tamil language. So that I and others can use it in the projects. 
- 75+ million Tamil speakers worldwide
- Official language in India, Sri Lanka, and Singapore
- Tamil Nadu is India’s 2nd largest IT hub, home to thousands of .NET developers
- Despite its global relevance, Tamil localization is currently missing in FluentValidation, even though many other .NET libraries already support it (e.g., ASP.NET Core, DataAnnotations, etc.)


JeremySkinner:
12.1.0 is now released with this change

Issue #2101, opened 2023-04-27
Rulesets and SetValidator


Robelind:
### FluentValidation version

11.5.2

### Summary

It seems that in certain scenarios when having rulesets with SetValidator(), the validations are not performed.

### Steps to Reproduce

Using the code below validation works as expected and you get 3 errors.
```
using FluentValidation;
using FluentValidation.Results;

IValidator<Test> validator = new TestValidator();
ValidationResult result = await validator.ValidateAsync(new Test(),
     strategy => strategy.IncludeRuleSets("Test", "Test2").IncludeRulesNotInRuleSet());

foreach(var error in result.Errors)
{
    Console.WriteLine(error.ErrorMessage);
}

Console.ReadKey();

public class TestValidator : AbstractValidator<Test>
{
    public TestValidator()
    {
        RuleFor(t => t.Value).Equal(56);
        RuleSet("Test", () => RuleFor(t => t).SetValidator(new TestValidator2()));
        RuleSet("Test2", () => RuleFor(t => t).SetValidator(new TestValidator3()));
    }
}

public class TestValidator2 : AbstractValidator<Test>
{
    public TestValidator2()
    {
        RuleFor(t => t.Value).Must(value => value == 12);
    }
}

public class TestValidator3 : AbstractValidator<Test>
{
    public TestValidator3()
    {
        RuleFor(t => t.Value).Must(value => value == 13);
    }
}

public class Test
{
    public int Value { get; set; }
}
```
If  we change the validation request to:
```ValidationResult result = await validator.ValidateAsync(new Test(), strategy => strategy.IncludeRuleSets("Test", "Test2"));```
 the rulsets doesn't seem to be executed and we get no validation errors.

Changing the implementation of `TestValidator2` to:
`RuleSet("Test", () => RuleFor(t => t).Must(t => t.Value == 12));`
yields one validation error.

Have I misunderstood something or is this intended behaviour?

**Comment by JeremySkinner** [comment]
This is the correct behaviour - ruleset selection cascades down to child validators. If you invoke the "Test" ruleset on your top-level validator and the child validator definition is inside the "Test" ruleset, then only rules inside the child validator also within a "Test" ruleset will be executed.

You can change/override the default behaviour by specifying a ruleset override in the call to SetValidator:

For example:

```
RuleSet("Test", () => { 
  RuleFor(t => t).SetValidator(new TestValidator(), "default")
});
```

[...]

Issue #2100, opened 2023-04-17
RulesetValidatorSelector does not pass Ruleset to child validator on collection


wonea:
### FluentValidation version

11.5.2

### ASP.NET version

11.3.0

### Summary

RulesetValidatorSelector does not pass Ruleset to child validator. The following should fail with the books not having a name.

### Steps to Reproduce

LINQPad 7 test, child objects should fail.

	void Main()
	{
		var requestObj = new Library("L3 8EW", "Libraries & Information Services", new List<Book>
		{
			new("9780545791434", null),
			new("9780439358071", null),
			new("9780545790352", null),
			new("9780747560722", null)
		});

		var serviceCollection = new ServiceCollection();
		serviceCollection.AddScoped<IValidator<Library>, LibraryValidator>();
		serviceCollection.AddScoped<IValidator<Book>, BookValidator>();
		var serviceProvider = serviceCollection.BuildServiceProvider();
		
		var serviceType = typeof(IValidator<Library>);

		if (serviceProvider.GetService(serviceType) is not IValidator validator)
		{
			return;
		}

		var selector = new RulesetValidatorSelector(new string[] { "Post" });
		// create validation context
		var validationContextGeneric = typeof(ValidationContext<>).MakeGenericType(typeof(Library));
		var validationContext = Activator.CreateInstance(validationContextGeneric, requestObj, new PropertyChain(),
			selector);

		// validate
		var result = validator.Validate((IValidationContext)validationContext);
		result.Dump();
		if (!result.IsValid)
		{
			throw new ValidationException(result.Errors);
		}
	}

	private class Library
	{
		public string PostCode { get; set; }
		public string Name { get; set; }

		public ICollection<Book> Books { get; set; }

		public Library() { }

		public Library(string postCode, string name, ICollection<Book> books)
		{
			PostCode = postCode;
			Name = name;
			Books = books;
		}

		public override string ToString()
		{
			return $"{nameof(PostCode)}: {PostCode}, {nameof(Name)}: {Name}, {nameof(Books)}: {Books.Count}";
		}
	}

	private class LibraryValidator : AbstractValidator<Library>
	{
		public LibraryValidator(IValidator<Book> bookValidator)
		{
			RuleFor(l => l.PostCode)
				.NotEmpty();

			RuleSet("Post", () =>
			{
				RuleFor(l => l.Name)
					.NotEmpty();
			});

			RuleForEach(l => l.Books)
[...]

PR #2337, opened 2025-08-27
docs: Document how RuleSets and SetValidator (child validators) work togeher

rklec:
I also stumbled upon this and it also seems to be common for others: https://github.com/FluentValidation/FluentValidation/issues/2101 https://github.com/FluentValidation/FluentValidation/issues/2100

So I guess at least document it, so people notice their problem faster.


JeremySkinner:
Thanks, I've slightly updated the wording and merged that.

PR #2339, opened 2025-09-30
Enable centralized package version management

AdelSS04:
Introduced `Directory.Packages.props` to centralize dependency version management using the `ManagePackageVersionsCentrally` property. Updated all project files to remove hardcoded package versions, relying on the centralized definitions for consistency. Removed redundant `<EnablePackageValidation>` entries and standardized project file formatting.

Updated specific project files to use centralized versions for dependencies like `BenchmarkDotNet`, `Bogus`, `Newtonsoft.Json`, `xunit`, and others. Added a conditional version for `System.Threading.Tasks.Extensions` targeting `netstandard2.0`.

These changes improve maintainability, ensure consistency, and simplify dependency management across the solution.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -1,0 +1,23 @@
+<Project>
+  <PropertyGroup>
+    <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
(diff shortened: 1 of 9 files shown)


JeremySkinner:
Thanks for the suggestion, but I'm happy with how things are currently and don't feel there's a need to introduce central package versioning.

PR #2354, opened 2026-03-26
Use sealed for private and internal classes.

kronic:


JeremySkinner:
I see no need for this

PR #2359, opened 2026-06-04
docs(ValidationFailure): Nullable annotations

FynZ:
Encountered an issue when collecting validation failures as you can have nullable property values which even tho is obvious was not highlighted by the nullable annotations for the property. Making this class nullable aware (at least partially, I believe `AttemptedValue` is the most accessed property which can be null).

Had to remove the default constructor since `PropertyName` can never be null, in theory `ErrorMessage` could be if overloaded with null but this would require propagating the nullable annotation upstream.

### DIFF
--- a/src/FluentValidation/Results/ValidationFailure.cs
+++ b/src/FluentValidation/Results/ValidationFailure.cs
@@ -21,18 +21,13 @@ namespace FluentValidation.Results;
 using System;
 using System.Collections.Generic;
 
+#nullable enable
+
 /// <summary>
 /// Defines a validation failure
 /// </summary>
 [Serializable]
 public class ValidationFailure {
-
-	/// <summary>
-	/// Creates a new validation failure.
-	/// </summary>
-	public ValidationFailure() {
-
-	}
 
 	/// <summary>
 	/// Creates a new validation failure.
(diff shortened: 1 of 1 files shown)


FynZ:
> @FynZ please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @dotnet-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @dotnet-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @dotnet-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

@dotnet-policy-service agree

JeremySkinner:
Thanks for the suggestion but this isn't something we'll be supporting

PR #2360, opened 2026-06-27
fix: support indexers in expression-based ShouldHaveValidationErrorFor

arashzjahangiri:
## Problem

`ShouldHaveValidationErrorFor(x => x.Orders[0].ProductName)` (expression overload) always failed to match, even when a validation error for `Orders[0].ProductName` existed. There were two independent bugs:

**Bug 1 — `PropertyChain.FromExpression` ignores indexers**

For `x => x.Orders[0].ProductName`, the walk stopped at the `MethodCallExpression` for `[0]` and returned only `"ProductName"` instead of `"Orders[0].ProductName"`.

**Bug 2 — Asymmetric normalization in `TestValidationResult`**

`ShouldHaveValidationError` strips indexers from the *actual* error PropertyName before comparing, but not from the *expected* propertyName resolved from the expression. After fixing Bug 1, the resolved name contains `[0]` while the normalized actual name does not — so they still fail to match.

**Bug 3 — `ArgumentNullException` on model-level rules** (uncovered by fixing Bug 2)

Passing both sides through `NormalizePropertyName` exposed a crash when `propertyName` is null (model-level rules where `PropertyName` resolves to null).

## Fix

- `PropertyChain.FromExpression`: detect `get_Item` indexer calls and append the index to the parent property segment (e.g. `"Orders[0]"`).
- `TestValidationResult`: normalize both sides — `NormalizePropertyName(x.PropertyName) == NormalizePropertyName(propertyName)`.
- `NormalizePropertyName`: guard against null input.

## Tests

- `PropertyChainTests.Creates_from_expression_with_indexer` — chain resolves `x => x.Orders[0].ProductName` to `"Orders[0].ProductName"`.
- `ValidatorTesterTester.Can_use_indexer_in_expression_message` — `ShouldHaveValidationErrorFor(x => x.Orders[0].ProductName)` finds the error.
- `ValidatorTesterTester.Can_use_indexer_in_expression_message_inverse` — `ShouldNotHaveValidationErrorFor` passes cleanly.

All 871 existing tests continue to pass.

> Addresses the TODO noted in #2057: "Need some tests around normalising property names for collection expressions in the test helper."

### DIFF
--- a/src/FluentValidation.Tests/PropertyChainTests.cs
+++ b/src/FluentValidation.Tests/PropertyChainTests.cs
@@ -84,6 +84,13 @@ public class PropertyChainTests {
 	}
 
 	[Fact]
+	public void Creates_from_expression_with_indexer() {
+		Expression<Func<Person, string>> expr = x => x.Orders[0].ProductName;
+		var chain = PropertyChain.FromExpression(expr);
+		chain.ToString().ShouldEqual("Orders[0].ProductName");
+	}
+
+	[Fact]
 	public void Should_ignore_blanks() {
 		chain.Add("");
 		chain.Add("Foo");
--- a/src/FluentValidation.Tests/ValidatorTesterTester.cs
+++ b/src/FluentValidation.Tests/ValidatorTesterTester.cs
@@ -713,6 +713,30 @@ public class ValidatorTesterTester {
 	}
 
 	[Fact]
+	public void Can_use_indexer_in_expression_message() {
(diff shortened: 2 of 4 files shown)


arashzjahangiri:
@dotnet-policy-service agree

JeremySkinner:
closing as duplicate of #2360, will copy your comments for bug 2/3 into that PR to address in future, if we decide to proceed with that feature. 

For now please use the string-based overload rather than the expression-based overload.

PR #2363, opened 2026-07-03
PropertyPath is incorrect in .When() on RuleForEach

faddiv:
I found this bug, but no place to report it. I also fixed it.

Problem: The `When` condition on the collection doesn't have the current property path, but previous or null, if the first condition.

The case which proves it: `Can_access_collection_item_property_path_in_condition`

Root cause: Since in `CollectionPropertyRule` the context is adjusted after the `When` condition is evaluated it contains the previous `PropertyPath`

Solution: Initialize the context, before the first use by external method calls.

Additional tests:
- ConditionTests has new tests that focus on the `PropertyPath` correctness.
- ValidatorSelectorTests has one additional test that focuses on the whole collection. (There were only sub-path tests.) I put the context initialization before the selector, because it also gets the context, but after testing, I realized it only uses the explicit `propertyPath` parameter. Regardless, I left it there.

### DIFF
--- a/src/FluentValidation.Tests/ConditionTests.cs
+++ b/src/FluentValidation.Tests/ConditionTests.cs
@@ -252,6 +252,73 @@ public class ConditionTests {
 		result.IsValid.ShouldBeFalse();
 	}
 
+	[Fact]
+	public void Can_access_property_path_in_condition() {
+		string propertyPath = null;
+		var validator = new TestValidator {
+			v => v.RuleFor(x => x.Forename).NotEmpty().When((_, c) => {
+				propertyPath = c.PropertyPath;
+				return true;
+			})
+		};
+
+		validator.Validate(new Person());
+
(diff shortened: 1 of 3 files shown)


faddiv:
@dotnet-policy-service agree

JeremySkinner:
I'm not actively working on this project atm, but if you're able to provide a fix I'll try to take a look at it when I get some time. Thanks!

faddiv:
@JeremySkinner I provided the fix.
Also, there are some additional tests:
- ConditionTests has new tests that focus on the `PropertyPath` correctness.
- ValidatorSelectorTests has one additional test that focuses on the whole collection. (There were only sub-path tests.) I put the context initialization before the selector, because it also gets the context, but after testing, I realized it only uses the explicit `propertyPath` parameter. Regardless, I left it there.

PR #2369, opened 2026-08-09
docs: Clarify RegularExpressionValidator anchoring (newline semantics)

OccasionalDJ:
Small PR to update docs and tests to clarify regex anchor behaviour, especially for anyone migrating from DataAnnotations.

Although this is standard .NET regex behaviour, think this is worth highlighting clearly as it can easily be overlooked by developers (and AI agents reading the docs/tests), leading to potentially incorrect validation being applied for typical single line field validations like the example "surname" test.

(The issue tracker isn't currently accepting new issues, so hope raising this PR without further discussion is ok)

### DIFF
--- a/docs/built-in-validators.md
+++ b/docs/built-in-validators.md
@@ -263,6 +263,14 @@ String format args:
 * `{RegularExpression}` – Regular expression that was not matched
 * `{PropertyPath}` - The full path of the property
 
+Note: If you're used to ASP.NET's `[RegularExpression]` DataAnnotations attribute, be careful not to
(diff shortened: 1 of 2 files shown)


OccasionalDJ:
@dotnet-policy-service agree

JeremySkinner:
thanks for this. Could you update the doc change to remove the final sentence, that references to client-side validation? The clientside validation integration is no longer supported, I just haven't got around to updating the doc page yet.

OccasionalDJ:
Hi @JeremySkinner - think this is good to merge now

Issue #2332, opened 2025-06-27
Issue tracker temporarily offline


JeremySkinner:
The issue tracker is temporarily offline as I'm not currently able to monitor it due to personal reasons. 

It will be re-enabled in the future. 

Thank you for your understanding.

PR #2371, opened 2026-09-21
fix: use GetValuesAsUnderlyingType so IsInEnum works under native AOT

Scrambles56:
### Problem

`IsInEnum()` on a `[Flags]` enum **throws at runtime** in a native-AOT published app:

```
Unhandled Exception: System.NotSupportedException: 'Access[]' is missing native code or metadata.
   at System.Array.CreateInstance(Type, Int32)
   at System.RuntimeType.GetEnumValues()
   at FluentValidation.Validators.EnumValidator`2.EvaluateFlagEnumValues(Int64, Type)
```

Minimal reproduction: https://github.com/Scrambles56/fluentvalidation-aot-repro (FluentValidation 12.1.1, reproduced on .NET 8, 9 and 10).

`EvaluateFlagEnumValues` calls `Enum.GetValues(Type)`, which is annotated `RequiresDynamicCode` because it constructs an array of the enum type at runtime — something AOT can't do unless that array type happened to be rooted. The publish warns about this (IL3050), but it's rolled up into an assembly-level `IL3053` by default, so it's easy to miss until the app crashes.

Only the `[Flags]` path is affected — non-flags enums go through `Enum.IsDefined`, which is AOT-safe.

### Change

```diff
-		foreach (var enumValue in Enum.GetValues(enumType)) {
+		foreach (var enumValue in Enum.GetValuesAsUnderlyingType(enumType)) {
```

`GetValuesAsUnderlyingType` returns the values already boxed as their underlying primitive, so no enum-typed array is constructed. The loop body is untouched — `Convert.ToInt64` handles the primitive identically, so behaviour is unchanged. It's available on .NET 8, which is this project's only target framework, so no conditional compilation is needed.

Incidentally it also removes a box/unbox round trip: the current code boxes each value as the enum type, then `Convert.ToInt64` immediately unboxes it.

### Verification

- The repro's IL3050 warning disappears and the published binary runs correctly
- All 22 existing `EnumValidatorTests` pass on net8.0, net9.0 and net10.0 — including the bitwise, overlapping-flags and zero-value cases
- Full suite: **877 passed, 0 failed, 1 skipped**

### Notes

I'm aware #2321 touched this same method and was declined as breaking. That PR was +110/−61 across 5 files, changed the `EnumValidator` API surface and added a new `NullableEnumValidator`. This is one line inside a `private static` method: no API change, no new types, no behaviour change.

CONTRIBUTING.md asks for an issue first — the tracker is currently closed (#2332), so I've opened this directly. Happy to add an explicit flags-enum regression test, or to close this if you'd rather handle it differently.


Scrambles56:
@dotnet-policy-service agree

